PR #16, opened 2021-01-09
Update gantt

jarrodmillman:


rossbar (approved):
Good idea, the Python support windows cause those "bars" to extend much further out, I agree a more compact timeline is better.

I am planning on generating these data in the mermaid gantt format from `versions.py`, but will do so in a later PR (unless someone beats me to it).

jarrodmillman:
@rossbar I did a first pass at generating the mermaid Gantt format from ``versions.py``.

PR #52, opened 2021-02-09
Add @carreau for IPython

Carreau:
.... and let see if the CI builds it

### DIFF
--- a/meta-spec-0001.md
+++ b/meta-spec-0001.md
@@ -34,9 +34,9 @@ They do, however,
 {{< gallery >}}
 
 {{< person
-      name="?"
-      github_handle=""
-      avatar=""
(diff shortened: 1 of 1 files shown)


meta-spec-0001.md
```diff
@@ -34,9 +34,9 @@ They do, however,
 {{< gallery >}}
 
 {{< person
-      name="?"
-      github_handle=""
-      avatar=""
+      name="Matthias Bussonnier"
+      github_handle="carreau"
+      avatar="https://avatars.githubusercontent.com/u/335567?s=200"
       projects="ipython"
```
jarrodmillman, line 40: We are bootstrapping by choosing people from projects.  But you should add any other relevant projects that you are working on separated by a comma.

PR #91, opened 2021-05-12
Add @juanis2112 to the steering committee

juanis2112:


Carreau:
it looks like you need to add a .DS_Store to your [global gitignore](https://docs.github.com/en/github/getting-started-with-github/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer)

juanis2112:
Thanks @Carreau, my bad. I will do that for next one.
Thanks @jarrodmillman for fixing it :)

PR #96, opened 2021-05-13
Update Lucy's info

jarrodmillman:


lucyleeow:
I do not know how I managed that - I probably forgot to save before I committed. Thanks - this looks fine.

PR #101, opened 2021-05-15
add Kira Evans to SC

kne42:


jarrodmillman:
@stefanv Kira's avatar isn't showing up in the deploy-preview.

steering-committee/kira_evans.md
```diff
@@ -0,0 +1,5 @@
+---
+title: Kira Evans
+repository: https://github.com/kne42
+avatar: https://twitter.com/kne42/photo
```
Carreau, line 4:
```suggestion
avatar: https://pbs.twimg.com/profile_images/1295752376003911680/w0-kdJ-p_400x400.jpg
```

The current URL is not a link to an image.

kne42: ah, good catch!

jarrodmillman:
Hmm.  The deploy-preview is still not showing the avatar picture.  The link to the jpg is in the page source, but the image does not show up.

Carreau:
> Hmm. The deploy-preview is still not showing the avatar picture. The link to the jpg is in the page source, but the image does not show up.

It does  for me:

Screen Shot 2021-05-15 at 10 16 29

Domain blocker ?

jarrodmillman:
It looks like it is my browser security settings.  Among other things I have "Enhanced  Tracking Protection" set to "on":
https://developer.mozilla.org/en-US/docs/Web/Privacy/Tracking_Protection

If I disable it, Kira's avatar is displayed.  Should we just merge this?  Thoughts?

jarrodmillman:
It looks like the behavior isn't the default (but something I selected).  So I tentatively think we should merge this as is.  If no one disagrees in the next hour or so, I will merge it later today.

Carreau:
Let's merge as is, and we can see about copying the assets in the repository at a later date.

PR #112, opened 2021-06-28
Add Paul to the list of IPython contacts.

Carreau:
Fernando would have loved but explicitly said he had no time.
There does not seem to be much time for participation from other as
well.

-- 
 
We also had a recent newcomers to the core committers of IPython – who joined after this project was started : @mrmino. I don't want to overwhelm them yet with extra responsibilities though

### DIFF
--- a/core-projects/ipython.md
+++ b/core-projects/ipython.md
@@ -1,6 +1,6 @@
 ---
 title: "IPython"
-draft: true
+draft: false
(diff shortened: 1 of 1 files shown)


jarrodmillman:
@Carreau Could you change
```
draft: true
```
to
```
draft: false
```

Thanks!

PR #130, opened 2022-03-01
Update SPEC-0001: refer to lazy_loader pypi package

stefanv:


tupui (approved):
LGTM. I suppose it's ok to not have a release not in RC until this is not a draft anymore.

PR #139, opened 2022-07-06
Add caveats about lazy loading SPEC001

tlambert03:
as requested by @jni over in https://github.com/scikit-image/scikit-image/pull/6429, this adds a section about the caveats of lazy loading with regards to IDEs and static type checkers.

### DIFF
--- a/spec-0001/index.md
+++ b/spec-0001/index.md
@@ -216,7 +216,7 @@ Static type checkers (such as [mypy](http://mypy-lang.org) and
 types and locations dynamically-loaded modules and functions. This means that some
 integrated development environments (e.g.
 [VS Code](https://code.visualstudio.com)) will not be able to provide code
-completion, parameter hints, documentation, or other features for any objects in your module.
+completion, parameter hints, documentation, or other features for any objects in your module.  Nor will `mypy` be able to detect potential errors in your user's code.
 
 You can direct static type checkers to the actual location of dynamically loaded objects in one of two ways:
 
--- a/spec-0001/index.md
+++ b/spec-0001/index.md
@@ -207,6 +207,65 @@ In the mean time, we now have the necessary mechanisms to implement it ourselves
 
 [^cannon]: Cannon B., personal communication, 7 January 2021.
 
+### Caveats
+
+While this pattern has benefits at runtime, it does have one specific drawback:
+
+Static type checkers (such as [mypy](http://mypy-lang.org) and
+[pyright](https://github.com/microsoft/pyright)) will not be able to infer the
+types and locations dynamically-loaded modules and functions. This means that some
+integrated development environments (e.g.
+[VS Code](https://code.visualstudio.com)) will not be able to provide code
+completion, parameter hints, documentation, or other features for any objects in your module.
+
+You can direct static type checkers to the actual location of dynamically loaded objects in one of two ways:
+
+1. Use an [`if
+   typing.TYPE_CHECKING`](https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING)
+   clause .
+
+   ```python
+    from typing import TYPE_CHECKING
+
+    import lazy_loader as lazy
+
+    if TYPE_CHECKING:
+        from .edges import sobel, sobel_h, sobel_v
+
+    __getattr__, __dir__, __all__ = lazy.attach(
+        __name__,
+        submod_attrs={
+            'edges': ['sobel', 'sobel_h', 'sobel_v']
+        }
+    )
+    ```
+
+    `typing.TYPE_CHECKING` is a special variable that is set to `False` at runtime, and has negative runtime impact. 
+
+2. Use a [type stub (`.pyi`) file](https://mypy.readthedocs.io/en/stable/stubs.html). Type stubs are ignored at runtime, but used
+by static type checkers. This method entails adding a new file with a `.pyi` extension in the same directory as your actual `.py`. For example:
+
+    ```python
+    # mypackage/__init__.py
+    import lazy_loader as lazy
+
+    __getattr__, __dir__, __all__ = lazy.attach(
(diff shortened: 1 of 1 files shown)


stefanv:
It strikes me that we have all the information available to generate classic imports. Would generating the stubs be helpful, or would it typically overwrite what is already there? Are there other ways in which we could guide the static type checkers without duplicating imports?

tlambert03:
> It strikes me that we have all the information available to generate classic imports. Would generating the stubs be helpful, or would it typically overwrite what is already there?

yep, for sure.  I can imagine (and could take a stab if you'd like) a utility in `lazy_loader` that would generate these stubs on demand (we have [something like this in napari](https://github.com/napari/napari/blob/main/napari/utils/stubgen.py) but this one would be much simpler).  The question then becomes when to "demand" it.  You could do it at test time in CI, to detect whether a change has been made to public exports.  (you could probably also add some sort of `pre-commit` thingy to do it automatically).  Or, you could run some stub-generation script when generating your sdist/wheel.  The choice between the two feels a bit subjective.  thoughts?

> Are there other ways in which we could guide the static type checkers without duplicating imports?

these are the only two I know of.  I know that they both feel unfortunate in terms of duplication.  I do generally implement some sort of automation/checking in my own projects, but I'm not sure I feel like there's an "obvious" one-size-fits all suggestion there

spec-0001/index.md
```diff
@@ -207,6 +207,64 @@ In the mean time, we now have the necessary mechanisms to implement it ourselves
...
+
+     __getattr__, __dir__, __all__ = lazy.attach(
+         __name__,
+         submod_attrs={
+             'edges': ['sobel', 'sobel_h', 'sobel_v']
+         }
+     )
+    ```
+
+    `typing.TYPE_CHECKING` is a special variable that is set to `False` at runtime, and has negative runtime impact.
```
jni, line 242:
```suggestion
    `typing.TYPE_CHECKING` is a special variable that is set to `False` at runtime, and has negligible runtime impact.
```

jni: Or "no". But not negative I presume! 😂

stefanv:
There is one alternative, and that is for the lazy loader to have a utility function that parses a pyi to generate its own imports. Either way, not sure we can do this perfectly cleanly without built-in support as per that recent PEP.

spec-0001/index.md
tlambert03: Lol, yes, negligible was the goal

tlambert03:
> and that is for the lazy loader to have a utility function that parses a pyi to generate its own imports.

good point!  how about this approach?  feels kinda clean and seems to work fine for (e.g.) `skimage.filters`

```python
# skimage/filters/__init__.py
from .._shared import lazy

__getattr__, __dir__, __all__ = lazy.attach_stub(__name__, __file__)
```

```python
# skimage/_shared/lazy.py
...

import ast

class Visitor(ast.NodeVisitor):
    def __init__(self) -> None:
        self._submodules = set()
        self._submod_attrs = {}

    def visit_ImportFrom(self, node: ast.ImportFrom):
        assert node.level == 1, 'Currently only support `import from .*`'
        if node.module:
            attrs: list = self._submod_attrs.setdefault(node.module, [])
            attrs.extend(alias.name for alias in node.names)
        else:
            self._submodules.update(alias.name for alias in node.names)


def attach_stub(package_name: str, filename: str):
    stubfile = f'{filename}i'

    if not os.path.exists(stubfile):
        raise ValueError(f"Stub file {stubfile} does not exist")

    with open(stubfile) as f:
        stub_node = ast.parse(f.read())

    visitor = Visitor()
    visitor.visit(stub_node)
    return attach(package_name, visitor._submodules, visitor._submod_attrs)
```

```python
# skimage/filters/__init__.pyi

from . import rank
from ._fft_based import butterworth
from ._gabor import gabor, gabor_kernel
from ._gaussian import difference_of_gaussians, gaussian
from ._median import median
from ._rank_order import rank_order
from ._sparse import correlate_sparse
from ._unsharp_mask import unsharp_mask
from ._window import window
from .edges import (farid, farid_h, farid_v, laplace, prewitt, prewitt_h,
                    prewitt_v, roberts, roberts_neg_diag, roberts_pos_diag,
                    scharr, scharr_h, scharr_v, sobel, sobel_h, sobel_v)
from .lpi_filter import LPIFilter2D, inverse, wiener
from .ridges import frangi, hessian, meijering, sato
from .thresholding import (apply_hysteresis_threshold, threshold_isodata,
                           threshold_li, threshold_local, threshold_mean,
                           threshold_minimum, threshold_multiotsu,
                           threshold_niblack, threshold_otsu,
                           threshold_sauvola, threshold_triangle,
                           threshold_yen, try_all_threshold)
```

jni:
I love that @tlambert03 😍

stefanv:
@tlambert03 I see the PR I requested already exists :heart_eyes:  Now the your lazy stubs PR has been merged, we can modify this one and get it in as well.  Thanks again!

tlambert03:
updated!  lemme know if you see anything missing, or would like anything changed

jni (approved):

stefanv:
@tlambert03 Sorry for the delay, I was on parental leave.  But now I'm back :)

I worked a bit on your PR to make it shorter, and to make the advice more direct (if opinionated).  Please take a look at https://github.com/tlambert03/specs/pull/1 and let me know what you think?

stefanv:
Thanks again for your work on the stubs loader!

stefanv:
@tlambert03 It looks like this trick doesn't quite fool `pyright`, which complains that the modules are not *actually* imported.

tlambert03:
Yes, I think there's no getting around a static, manually entered `__all__` in that case.  This approach isn't incompatible with that, but that would be an additional step

PR #157, opened 2022-09-23
Add TOC to README

pllim:
I do not want to click on each and every folder to see what the SPECs are about, so would be nice to have a TOC like https://github.com/astropy/astropy-APEs . 😅 

I am not sure if these are the columns you want, so I am happy to modify the columns. Would be nice for outsider to see the statuses of the SPECs at a glance too. And once SPEC is accepted, would be nice to have a permanent DOI people can cite.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,3 +4,14 @@ The Scientific Python Ecosystem Coordination (SPEC) mechanism is used to recomme
 project policies, coding conventions, and standard tooling.
 
 See https://scientific-python.org/specs/
+
+
+| # | Title | Proposed | Accepted | DOI |
+| --- | --- | --- | --- | --- |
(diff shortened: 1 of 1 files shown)


jarrodmillman:
This is auto-generated: https://scientific-python.org/specs/ and has all the information in the README table, but will also automatically include additional information that may be hard to include in the README (for example which projects have endorsed the SPEC and what its status is).

I would rather not have too much additional manual work needed. I also don't like having to manually format tables. Is the website not sufficient? If not, maybe we should consider improving the `quickstart.py` script to also update the `README` using `tabulate` or something.

pllim:
Okay, fair point. I will close this.

PR #158, opened 2022-09-28
Discuss explicit import exports for stub loader

stefanv:


tlambert03:
Is pyright also happy with declaring `__all__` in the stub (rather than using X as X for every item?). If so, I think it's worth mentioning

jarrodmillman (approved):

PR #168, opened 2022-12-20
Add SPEC 6: Keys to the Castle

stefanv:
/cc @jarrodmillman

I think the gopass implementation is user friendly enough that we can use it as an example.

Do you have any other ideas for principles the system should adhere to?

### DIFF
--- a/spec-0006/index.md
+++ b/spec-0006/index.md
@@ -1,0 +1,66 @@
+---
+title: "SPEC 6 — Keys to the Castle"
+date: 2022-12-19
+author:
+  - "Stéfan van der Walt <stefanv@berkeley.edu>"
+  - "Jarrod Millman <millman@berkeley.edu>"
+discussion:
+endorsed-by:
+---
+
+## Description
+
+<!--
+Briefly and clearly describe the proposal.
+Explain the general need and the advantages of this specific proposal.
+If relevant, include examples of how the new functionality would be used,
+intended use-cases, and pseudo-code illustrating its use.
+-->
+
+Project developers often have to share secrets, such as server passwords and social media logins.
+Typically, such secrets are distributed among those who need them, without any centralized system for tracking the secrets or who has access to them.
+When a server needs to be accessed, there is often a scramble to find someone with credentials.
+This SPEC discusses the requirements for a system to distribute secrets, and provides an example implementation.
+
+## Implementation
+
+### Principles
+
+A system for distributing project secrets must have the following properties:
+
+- Secrets are stored encrypted in a central (remote) location.
+- It must be possible to grant access to the secrets to a select group of team members.
+- It must be possible to revoke future access to the secrets.
+- The system must not rely on closed source or commercial encryption facilities, that
+  can later disappear or be made unavailable.
+
+### Example
+
+The [NumPy password vault](https://github.com/numpy/vault) is an example of an implementation that satisfies the above principles.
+The secrets are stored, encrypted, in a public Git repository.
+The vault uses [gopass](https://github.com/gopasspw/gopass) to manage access via GPG keys.
+Each secret is encrypted using the public keys of all developers that should have access.
+If a developer's access is removed, the vault is re-encrypted so that that developer cannot read future copies of the repository.
+
+### Alternative implementations
+
+Unknown
+
+### Core Project Endorsement
+
+<!--
+Discuss what it means for a core project to endorse this SPEC.
+-->
+
+### Ecosystem Adoption
+
+<!--
+Discuss what it means for a project to adopt this SPEC.
+-->
+
+## Notes
+
+<!--
+Include a bulleted list of annotated links, comments,
+and other ancillary information as needed.
+-->
--- a/spec-0006/index.md
+++ b/spec-0006/index.md
@@ -17,38 +17,48 @@ If relevant, include examples of how the new functionality would be used,
 intended use-cases, and pseudo-code illustrating its use.
 -->
 
-Project developers often have to share secrets, such as server passwords and social media logins.
+Projects engage with restricted resources all the time.
+Examples include access to add & remove team members, grant commit rights, or make uploads to certain hosts.
+
+**Documenting resources** is critical to ensuring uninterrupted operations.
+It is important that team members know _who_ have access to resources, and how to _gain access_ to resources.
+For example, if updated project documentation needs to be uploaded to a remote server via SSH, who are the team members that can do that, and by which process can a release manager request access?
+
+Furthermore, project developers sometimes have to **share secrets**, such as server passwords and social media logins.
 Typically, such secrets are distributed among those who need them, without any centralized system for tracking the secrets or who has access to them.
 When a server needs to be accessed, there is often a scramble to find someone with credentials.
+
 This SPEC discusses the requirements for a system to distribute secrets, provides an example implementation, and lists suitable hosted services.
+
+It should be noted that, in many cases, secrets & passwords can be avoided.
+Most online services (GitHub, PyPi, etc.), have permissioning systems through which developers can be given the required access.
+Access to servers can be given through SSH keys, which each developer has to safeguard.
+This document deals with the situation in which that is not possible, and secrets have to be shared.
 
 ## Implementation
 
 ### Principles
 
-A system for distributing project secrets must have the following properties:
+1. Restricted project resources must be documented.
 
-- Secrets are stored encrypted in a central (remote) location.
-- It must be possible to grant access to the secrets to a select group of team members.
-- It must be possible to revoke future access to the secrets.
-- The system must not rely on closed source or commercial encryption facilities, that
-  can later disappear or be made unavailable, although such a solution can be considered if it allows for all data to be exported.
+2. A system for distributing project secrets must have the following properties:
 
-_Tokens_ form a common subset of secrets stored. These should always be scoped (minimal necessary permissions), and set to expire after a reasonable time period (a year, typically). Instructions for rotating and revoking the token should be saved alongside it.
+   - Secrets are stored encrypted in a central (remote) location.
+   - It must be possible to grant access to the secrets to a select group of team members.
+   - It must be possible to revoke future access to the secrets.
+   - The system must not rely on closed source or commercial encryption facilities, that
+     can later disappear or be made unavailable, although such a solution can be considered if it allows for all data to be exported.
 
-Whichever system is chosen, its user interface should match the capabilities of the intended users.
-This reduces the risk of passwords being copied out to less secure mechanisms such as sticky notes or text files.
-If the target audience is not used to GPG or the command prompt, for example, the `pass` implementation below may not work.
+   Whichever system is chosen, its user interface should match the capabilities of the intended users.
+   This reduces the risk of passwords being copied out to less secure mechanisms such as sticky notes or text files.
+   If the target audience is not used to GPG or the command prompt, for example, the `pass` implementation below may not work, and an alternative like 1password or bitwarden should be considered.
 
-### Example implementation: pass
+3. _Tokens_ form a common subset of secrets stored. These should always be scoped (minimal necessary permissions), and set to expire after a reasonable time period (a year, typically).
+   Instructions for rotating and revoking the token should be documented.
 
-The [NumPy password vault](https://github.com/scientific-python/vault-template) is an example of an implementation that satisfies the above principles.
-The secrets are stored, encrypted, in a public Git repository.
-The vault uses [gopass](https://github.com/gopasspw/gopass), a more user friendly implementation of [pass](https://www.passwordstore.org/), to manage access via GPG keys.
-Each secret is encrypted using the public keys of all developers that should have access.
-If a developer's access is removed, the vault is re-encrypted so that that developer cannot read future copies of the repository (but secrets should be considered compromised and, thus, rotated).
+### Hosted password storage
 
-### Alternatives
+The following hosted solutions conform to the principles in (2) above.
 
 Sponsored solutions:
 
@@ -61,6 +71,14 @@ Self-hosted solutions:
 Paid solutions:
 
 - [bitwarden](https://bitwarden.com/) provides a non-profit discount of 25%
+
+### Offline password storage: pass
+
+[password vault](https://github.com/scientific-python/vault-template) is an example of an implementation that satisfies the above principles.
+The secrets are stored, encrypted, in a public Git repository.
+The vault uses [gopass](https://github.com/gopasspw/gopass), a more user friendly implementation of [pass](https://www.passwordstore.org/), to manage access via GPG keys.
+Each secret is encrypted using the public keys of all developers that should have access.
+If a developer's access is removed, the vault is re-encrypted so that that developer cannot read future copies of the repository (but secrets should be considered compromised and, thus, rotated).
 
 ### Other security considerations
 
--- a/spec-0006/index.md
+++ b/spec-0006/index.md
@@ -36,11 +36,11 @@ A system for distributing project secrets must have the following properties:
 
 ### Example
 
-The [NumPy password vault](https://github.com/numpy/vault) is an example of an implementation that satisfies the above principles.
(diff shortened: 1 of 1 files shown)


spec-0006/index.md
```diff
@@ -0,0 +1,68 @@
...
+Project developers often have to share secrets, such as server passwords and social media logins.
+Typically, such secrets are distributed among those who need them, without any centralized system for tracking the secrets or who has access to them.
+When a server needs to be accessed, there is often a scramble to find someone with credentials.
+This SPEC discusses the requirements for a system to distribute secrets, and provides an example implementation.
+
+## Implementation
+
+### Principles
+
+A system for distributing project secrets must have the following properties:
```
tupui, line 29: I would also add that everything must be documented. E.g. if you add a key, it should at minima say: scope, how to update/revoke/rotate, expiry, comment.

spec-0006/index.md
```diff
@@ -0,0 +1,68 @@
...
+
+- Secrets are stored encrypted in a central (remote) location.
```
tupui, line 31: I would strongly advice the use of a 2FA hardware based security key. Credentials of popular packages are too important to not use the most secure protocols.

tupui (commented):
Good idea 👍 I have some suggestions.

betatim:
In general I think this is a good SPEC. For mybinder.org we have been doing essentially this using `git-crypt` and one shared secret. This has worked well.

One "big picture" thing I wanted to point out, despite not having a good solution, is that command line tools like gopass and git-crypt don't have the same UX comfort factor as something with a GUI. Especially for people who don't spend all day using the terminal. This means for people like that you need to spend more time helping them, increase the chances of them storing the passwords in an additional location that is easier to access, increase the chance of being turned off because the friction is too high, etc. So I wonder if there is a tool that has a nice GUI that can be recommended. Or if there are password vault services that offer a free tier that allows sharing. In the past Keybase had a "not terrible" UX but their funding situation changed a few years ago (I think got bought by Zoom), which is a bit of a turn off. I don't think Firefox's password vault supports sharing, but maybe there is something to explore there (with Mozilla)?

tupui:
Good points 👍 AFAIK there are 2 viable options (OSS you can self-host) in terms of UI with sharing features (things like keepass are great for local use but not much for working with a team): Bitwarden and passbolt.

In both cases, the difficulty will be trust. Either you trust their system and use their tiers (possible for big projects as we have funds), or you self-host. But unless you have a security expert in the team, I would personally have reservations.

For some reason, we are not so much targeted by hackers yet. But once they realise we would be a great vector of attack, things would drastically change. So far it's just typo squatting and a bit of malware sharing on PyPi. But they could very easily target people with credentials to ship compromised binaries... And that's only the the easy quick thing someone could do. If you follow a bit the security and pen. test space, real life hacks are really crazy now (e.g. how they compromised LastPass.)

MridulS:
+1 on having a more user friendly solution for sharing secrets, currently need to pass around the git-crypt one and it just doesn't seem right. It's possible to self host Bitwarden (either with the official one or vaultwarden) but even paying for Bitwarden turns out not to be super expensive (we can just use the "family" plan). 

Also a +1 on 2FA hardware based security key as part of the SPEC. PyPI already "forced" people to have 2FA so I think we can elevate that requirement as a SPEC. I'm not sure if conda-forge feedstocks force people to have 2FA enabled on github or not but having a general guideline like this would be great!

tupui:
We could probably get free accounts from Bitwarden if we pitch correctly the idea.

stefanv:
The goal of the SPEC is to define principles, and any system that satisfies those principles is fine. We can give multiple examples.

We obviously need at least one free implementation (which I provided), and self-hosting is not typically an option for most projects. In fact, most projects don't even have funds to pay for bitwarden, and the account needs to be accessible by all team members.

spec-0006/index.md
```diff
@@ -0,0 +1,68 @@
...
+- The system must not rely on closed source or commercial encryption facilities, that
+  can later disappear or be made unavailable.
+
+### Example
+
+The [NumPy password vault](https://github.com/numpy/vault) is an example of an implementation that satisfies the above principles.
+The secrets are stored, encrypted, in a public Git repository.
+The vault uses [gopass](https://github.com/gopasspw/gopass) to manage access via GPG keys.
+Each secret is encrypted using the public keys of all developers that should have access.
+If a developer's access is removed, the vault is re-encrypted so that that developer cannot read future copies of the repository.
```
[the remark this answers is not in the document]
stefanv: That's a good idea, if the departure is not amicable.

spec-0006/index.md
```diff
@@ -0,0 +1,68 @@
...
+
+A system for distributing project secrets must have the following properties:
+
+- Secrets are stored encrypted in a central (remote) location.
+- It must be possible to grant access to the secrets to a select group of team members.
+- It must be possible to revoke future access to the secrets.
+- The system must not rely on closed source or commercial encryption facilities, that
+  can later disappear or be made unavailable.
+
+### Example
```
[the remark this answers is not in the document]
stefanv: We didn't implement anything, btw, this is a standard solution based on standard tools (Unix Password and GPG).

spec-0006/index.md
```diff
@@ -0,0 +1,68 @@
...
+This SPEC discusses the requirements for a system to distribute secrets, and provides an example implementation.
+
+## Implementation
+
+### Principles
+
+A system for distributing project secrets must have the following properties:
+
+- Secrets are stored encrypted in a central (remote) location.
+- It must be possible to grant access to the secrets to a select group of team members.
```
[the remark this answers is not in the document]
stefanv: I think this requirement would be too honerous; a person could create multiple vaults if you really need to separate concerns to that degree.

spec-0006/index.md
```diff
@@ -0,0 +1,68 @@
...
+- It must be possible to revoke future access to the secrets.
```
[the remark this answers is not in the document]
stefanv: +1

tupui:
> We obviously need at least one free implementation (which I provided)

@stefanv Note that https://github.com/numpy/vault is not publicly available. I don't have access to see the repo.

spec-0006/index.md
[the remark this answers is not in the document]
stefanv: Added a note on that.

spec-0006/index.md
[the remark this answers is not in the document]
stefanv: > > We obviously need at least one free implementation (which I provided)
> 
> @stefanv Note that https://github.com/numpy/vault is not publicly available. I don't have access to see the repo.

Thanks for pointing that out; I've now put a publicly visible repository in place and updated the link accordingly.

stefanv:
I will write to Bitwarden to find out about community pricing.

stefanv:
> > We obviously need at least one free implementation (which I provided)
> 
> @stefanv Note that https://github.com/numpy/vault is not publicly available. I don't have access to see the repo.

New link: https://github.com/scientific-python/vault-template

stefanv:
> I will write to Bitwarden to find out about community pricing.

> Bitwarden offers a 25% discount on our advertised pricing to registered non-profit organizations. If you sign up, you will start with a 7-day trial. At that time, let us know the name of your Organization, and we will apply the discount code to your account so that the discount will be applied to your first invoice.

stefanv:
See https://github.com/dani-garcia/vaultwarden

betatim:
> The goal of the SPEC is to define principles, and any system that satisfies those principles is fine. We can give multiple examples.

What I was hoping for is to get the "When setting things up, consider the UX comfort for all the different types of people involved. You don't want to create the equivalent of super-secure-password-complexity-and-rotation-rules that lead to people writing the password on a post-it stuck to the monitor".

spec-0006/index.md
```diff
@@ -0,0 +1,68 @@
...
+
+The [NumPy password vault](https://github.com/scientific-python/vault-template) is an example of an implementation that satisfies the above principles.
+The secrets are stored, encrypted, in a public Git repository.
+The vault uses [gopass](https://github.com/gopasspw/gopass) to manage access via GPG keys.
+Each secret is encrypted using the public keys of all developers that should have access.
+If a developer's access is removed, the vault is re-encrypted so that that developer cannot read future copies of the repository (but secrets should be considered compromised and, thus, rotated).
+
+### Alternative implementations
+
+Unknown
```
ivirshup, line 47: I was trying to figure out some [PyPI security stuff](https://discuss.python.org/t/pypi-security-work-multifactor-auth-progress-help-needed/1042/53) where 1password's "free team for open source" was recommended. It looks like it's quite popular with some large projects like: numfocus, jupyter, Python Software Foundation Infrastructure Team, and Rust.

https://github.com/1Password/1password-teams-open-source

I think a big advantage here is access controls, which I'm not sure you can get for free with bitwarden.

tupui: oh thanks for sharing that! I even see Kali in the list. If that's not a good sign, I don't know what would be 😅

tupui: Actually I just heard that NumFOCUS provide team accounts for affiliated projects.

jni: The NumFOCUS folks just point you to that 1Password page 😅 but yes, we are using 1Password for Teams for Open Source for napari and it was a [smooth process](https://github.com/1Password/1password-teams-open-source/pull/708).

tupui: Oh really? I heard a different story from another project. That would be a bit underwhelming but ok still nice to have a central place saying what exists and is possible to do.

ivirshup: @jni, could you add all the core devs to the team? It's been a little unclear to me if the team accounts were capped at 10.

jni: Well, not all 14 core devs have asked for invites so we haven't tested that hypothesis yet 😂 , but nothing in the process suggested that limit. They explicitly add for the number of core contributors in the PR template, so I presume that's the limit they set? But probably can be expanded?

stefanv: This is a great option; thanks for finding it! I've added it to the SPEC.

stefanv:
The document has been updated with the feedback above; thanks everyone!

spec-0006/index.md
ivirshup: 1password's [pricing page](https://1password.com/business-pricing) says the "teams" account is up to ten people, and after that it's called "business".

But not sure what this translates to for the open source projects.

spec-0006/index.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Description
+
+<!--
+Briefly and clearly describe the proposal.
+Explain the general need and the advantages of this specific proposal.
+If relevant, include examples of how the new functionality would be used,
+intended use-cases, and pseudo-code illustrating its use.
+-->
+
```
rgommers, line 19: @stefanv can you add that the starting principle is: don't use passwords unless you absolutely cannot avoid it? Servers should use SSH keys, PyPI, GitHub etc. all have concepts of Teams, etc.

stefanv:
I've widened the scope of the SPEC slightly (adding that projects should document who has access to which resources), made the note on scoped tokens / rotation more prominent, and moved the pass-based solution down as a second option.

stefanv:
> I didn't follow the entire convo so sorry if I missed it. Was https://blog.pypi.org/posts/2023-04-20-introducing-trusted-publishers/ considered?

We use trusted publishing, but that doesn't alter the fact that we sometimes need to store and share credentials.

stefanv:
> > I didn't follow the entire convo so sorry if I missed it. Was https://blog.pypi.org/posts/2023-04-20-introducing-trusted-publishers/ considered?
> 
> We use trusted publishing, but that doesn't alter the fact that we sometimes need to store and share credentials.

@pllim I thought about your idea a bit more, and perhaps it's worth mentioning some of these commonly used mechanisms explicitly. I'll add a note to the PR.

spec-0006/index.md
```diff
@@ -0,0 +1,113 @@
...
+The secrets are stored, encrypted, in a public Git repository.
+The vault uses [gopass](https://github.com/gopasspw/gopass), a more user friendly implementation of [pass](https://www.passwordstore.org/), to manage access via GPG keys.
+Each secret is encrypted using the public keys of all developers that should have access.
+If a developer's access is removed, the vault is re-encrypted so that that developer cannot read future copies of the repository (but secrets should be considered compromised and, thus, rotated).
+
+### Other common scenarios
+
+- **Publishing packages**: PyPi provides a [trusted publisher](https://docs.pypi.org/trusted-publishers/using-a-publisher/) mechanism for avoiding passwords
+
+### Other security considerations
```
lagru, line 89: This could also list a bullet point on what makes a secure secret in case of conventional passwords (e.g passphrases, minimal password length, etc). I imagine using the recommended password managers here, makes it easy to use long ones that don't have to be remembered and can just be copy-pasted. Or maybe just point out this fact.

Might be to detailed for the scope of this document though...

spec-0006/index.md
```diff
@@ -0,0 +1,113 @@
...
+This SPEC discusses the requirements for a system to distribute secrets, provides an example implementation, and lists suitable hosted services.
+
+It should be noted that, in many cases, secrets & passwords can be avoided.
+Most online services (GitHub, PyPi, etc.), have permissioning systems through which developers can be given the required access.
+Access to servers can be given through SSH keys, which each developer has to safeguard.
+This document deals with the situation in which that is not possible, and secrets have to be shared.
+
+## Implementation
+
+### Principles
```
lagru, line 40: Nitpick: Feels weird to me that "Principles" is in the section "Implementation". I'd make it its own level-`##` section after "Description" and before "Implementation".

tupui (commented):
I had another look 😃 I think if we want to move this forward it might be good to release. Then it will get more traction (e.g. as seen on SPEC4) and people will start to engage.

spec-0006/index.md
```diff
@@ -0,0 +1,113 @@
...
+
+Furthermore, project developers sometimes have to **share secrets**, such as server passwords and social media logins.
+Typically, such secrets are distributed among those who need them, without any centralized system for tracking the secrets or who has access to them.
+When a server needs to be accessed, there is often a scramble to find someone with credentials.
+
+This SPEC discusses the requirements for a system to distribute secrets, provides an example implementation, and lists suitable hosted services.
+
+It should be noted that, in many cases, secrets & passwords can be avoided.
+Most online services (GitHub, PyPi, etc.), have permissioning systems through which developers can be given the required access.
+Access to servers can be given through SSH keys, which each developer has to safeguard.
```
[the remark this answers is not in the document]
stefanv: Note passkeys are now an entirely new thing :)

spec-0006/index.md
```diff
@@ -0,0 +1,113 @@
...
+Each secret is encrypted using the public keys of all developers that should have access.
+If a developer's access is removed, the vault is re-encrypted so that that developer cannot read future copies of the repository (but secrets should be considered compromised and, thus, rotated).
+
+### Other common scenarios
+
+- **Publishing packages**: PyPi provides a [trusted publisher](https://docs.pypi.org/trusted-publishers/using-a-publisher/) mechanism for avoiding passwords
+
+### Other security considerations
+
+- **2FA**: Where possible, developers should use two-factor authentication for service logins.
```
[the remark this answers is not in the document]
stefanv: "Must" is not something a SPEC has any ability to enforce, but I can remove the "where possible".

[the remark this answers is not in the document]
stefanv: Given that I already use must elsewhere, I simply used it here too :)

stefanv:
I've addressed the feedback by @lagru and @tupui—thank you both!

jarrodmillman (approved):
There is a minor section rearrangement needed, but I will do that in a follow-up PR. Let's just merge this and then revise it as needed.

tupui (approved):
Thanks @stefanv 👏 I think this is great and now if projects do follow these guidelines that would be a huge step up in terms of security.

stefanv:
Great, and I'm happy to incorporate further suggestions. I found it quite helpful to aggregate these.

PR #194, opened 2023-05-24
[Experiement] virtual simple index.

Carreau:
Having to publish all wheels in the same repository is complicated from the size of the storage and token regenration. Any changes to the index URL needs update to all the CI.

So the idea is to have a virtual index that just point to the original one(s) that we can manipulate the way we want:

### DIFF
--- a/virtindex.py
+++ b/virtindex.py
@@ -1,0 +1,73 @@
+# using requests get the html from https://pypi.anaconda.org/scipy-wheels-nightly/simple/matplotlib/
+
+
+import requests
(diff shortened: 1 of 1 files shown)


Carreau:
Not sure this is the best place, but that may affect spec 4 and 5

Carreau:
it does not support the top /simple/ that list all the packages, nor the data-python-requires. 

It appears also the Pypi now expose wheel metadata w/o downloading them, but i'm unsure how.

PR #197, opened 2023-05-26
Update spec0 with extended dates

munkm:
This PR updates the index.md file to include:
* dates out to 2026 for python support
* new quarterly deprecation cycle drop support printing


munkm:
The linter is failing because of the indentation of packages in the quarterly release cycle. I think we want to keep those indentations. What do you think @Carreau ?

bsipocz (approved):
I think we should keep the indents, maybe the just need a `-` or similar to make it nested lists?

jarrodmillman (approved):

bsipocz (changes requested):
Rendering is not correct yet, so I'm revoking the green checkmark to avoid accidental premature merge

munkm:
Ok @bsipocz I fixed the formatting. What do you think now? 

@Carreau I made everything a sub-list. I didn't update your formatting script because I figured stuff might change as we automate, but I can add it if you'd like.

bsipocz (approved):
Looks all good now, thanks!

PR #216, opened 2023-05-31
SPEC4: change permission for token

tupui:
Just did this for SciPy and I needed `api:read` when uploading the first package. (I correctly deleted the token once the package was up.)


tupui:
mmm so apparently to create the package we need read, but this is not necessary for uploading new versions. Probably ok to advise to use read for everyone. Let me know if I should update and make the distinction.

jarrodmillman (approved):

PR #250, opened 2023-07-25
Avoid import aliasing in SPEC 1

stefanv:
See https://github.com/scientific-python/specs/pull/245#issuecomment-1650617352


rgommers (approved):
Thanks Stefan!

bsipocz (approved):

PR #255, opened 2023-08-02
NumPy project endorses SPEC 4

InessaPawson:
After the extensive review and discussion, the NumPy Steering Council came to the consensus on endorsing SPEC 4.


tupui (approved):
yay thank you 🎉

PR #269, opened 2023-09-27
Make the dependencies part explicit

max-sixty:
Clarify that the recommendation is about "how long new releases of a package should permit older versions of dependencies", not the more conventional meaning of support as "we fix bugs for old versions" — like most medium-sized open-source projects, we (xarray) only release fixes for the current version of xarray

ref https://github.com/pydata/xarray/discussions/8237#discussioncomment-7126175

@andersy005 @shoyer @dcherian (added by @jarrodmillman )

### DIFF
--- a/spec-0000/index.md
+++ b/spec-0000/index.md
@@ -29,8 +29,8 @@ All versions refer to feature releases (i.e., Python 3.8.0, NumPy 1.19.0; not Py
 
 Specifically, we recommend that:
 
-1. Support for a given version of Python be dropped **3 years** after its initial release.
-2. Support for a given version of other core packages be dropped **2 years** after their initial release.
+1. Support for Python versions be dropped **3 years** after its initial release.
+2. Support for core package dependencies be dropped **2 years** after their initial release.
 
 ### Core Project Endorsement
 
--- a/spec-0000/index.md
+++ b/spec-0000/index.md
@@ -29,7 +29,7 @@ All versions refer to feature releases (i.e., Python 3.8.0, NumPy 1.19.0; not Py
 
 Specifically, we recommend that:
 
-1. Support for Python versions be dropped **3 years** after its initial release.
+1. Support for Python versions be dropped **3 years** after their initial release.
 2. Support for core package dependencies be dropped **2 years** after their initial release.
 
 ### Core Project Endorsement


spec-0000/index.md
```diff
@@ -29,8 +29,8 @@ All versions refer to feature releases (i.e., Python 3.8.0, NumPy 1.19.0; not Py
 
 Specifically, we recommend that:
 
-1. Support for a given version of Python be dropped **3 years** after its initial release.
-2. Support for a given version of other core packages be dropped **2 years** after their initial release.
+1. Support for Python versions be dropped **3 years** after its initial release.
+2. Support for core package dependencies be dropped **2 years** after their initial release.
```
stefanv, line 33: Thanks @max-sixty!

I read the other issue, and I wonder if the following wouldn't more accurately describe what you discussed there:

```suggestion
1. Support for a given version of Python be dropped no later than **3 years** after its initial release.
2. Support for a given version of other core packages be dropped no later than **2 years** after their initial release.
```

stefanv, line 33: Thanks @max-sixty! I am trying to distill the intent of the conversation on the linked issue. I think what we're trying to say is:

(1) Limit maintenance burden, you don't need to support packages for so long.
(2) But be mindful that, if you don't support them roughly this length, other packages may struggle to install/use your library.

Maybe something like:

```suggestion
1. Support for a given Python version be dropped at latest **3 years** after its initial release.
2. Support for a given version of other core packages be dropped at latest **2 years** after their initial release.

This defines a maximum level of support, and while they may choose a much shorter window, doing so will make it harder to install their package along with others. 
```

@jarrodmillman

stefanv: Also not sure we want to even prescribe "at latest"; libraries can fully choose how long they maintain. This is almost a guidepost for where you want to aim, and beyond which you should have no guilt in dropping support.

max-sixty: Thanks @stefanv 

I had this as two changes
- Clarify "support" — the original intent of this PR
- Understand whether "two years" is a suggestion / maximum / recommendation? It seems like this isn't clear and could use some more consensus building? I thought it was a good idea to have a standard, so thank you for proposing one. If the standard is just a maximum, then the standard is less useful — for example, xarray depends on pandas, so there's no point xarray taking on two years if pandas were to select one year. Maybe we need to all agree on two years, even if that's a bit long for xarray & networkx? Or maybe the standard is "two years, for inner-core packages such as pandas / scipy / matplotlib, and one year for other packages"?

jarrodmillman: These are recommendations. So I am not sure how much detail we should provide about how you might not always follow the recommendations. I would slightly prefer to keep the recommendation that you drop it after two years.

With the understanding like for all good recommendations, that you should follow them as much as possible. But there will be times when you won't follow the recommendation. But ideally the exceptions should be minimal.

max-sixty: I think that's very reasonable

jarrodmillman (approved):

spec-0000/index.md
```diff
@@ -29,8 +29,8 @@ All versions refer to feature releases (i.e., Python 3.8.0, NumPy 1.19.0; not Py
 
 Specifically, we recommend that:
 
-1. Support for a given version of Python be dropped **3 years** after its initial release.
-2. Support for a given version of other core packages be dropped **2 years** after their initial release.
+1. Support for Python versions be dropped **3 years** after its initial release.
```
jarrodmillman, line 32:
```suggestion
1. Support for Python versions be dropped **3 years** after their initial release.
```

PR #271, opened 2023-10-01
add MNE to list of SPEC001 adopters

drammock:
Since MNE-Python isn't a core project, it isn't possible for us to endorse any SPECs. So this is a question-as-PR: do you want a record of all the non-core ecosystem projects that *adopt* each SPEC?  If not feel free to close.


jarrodmillman:
It is probably worth thinking about how to handle adopters now. Thanks for raising the issue. I am approving this for now and will merge, since the information may be useful for other projects. But we probably want to have something more general going forward. I will bring it up at the next SPEC meeting.

jarrodmillman (approved):

PR #272, opened 2023-10-02
Clarify schedule is guideline

Carreau:
I think otherwise the spec page seem a bit harsh.

### DIFF
--- a/spec-0000/index.md
+++ b/spec-0000/index.md
@@ -73,10 +73,7 @@ The situation has since improved due to improved installations via binary wheels
 Below is an auto generated schedule with recommended dates for dropping support. We suggest that the next release in a given quarter is
 considered as the one removing support for a given item.
 
-In particular for Python releases, as all releases are around October, you may
-want to delay removal of support of an older Python version until your package
-work on the newly released Python version, thus keeping the number of minor
-version of python supported the same for all releases.
+"you may want to delay the removal of support of an older Python version until your package fully works on the newly released Python, thus keeping the number of supported minor versions of Python the same for your package."
 
 {{< include-md "schedule.md" >}}
 
--- a/spec-0000/index.md
+++ b/spec-0000/index.md
@@ -73,7 +73,7 @@ The situation has since improved due to improved installations via binary wheels
 Below is an auto generated schedule with recommended dates for dropping support. We suggest that the next release in a given quarter is
 considered as the one removing support for a given item.
 
-"you may want to delay the removal of support of an older Python version until your package fully works on the newly released Python, thus keeping the number of supported minor versions of Python the same for your package."
+You may want to delay the removal of support of an older Python version until your package fully works on the newly released Python, thus keeping the number of supported minor versions of Python the same for your package.
 
 {{< include-md "schedule.md" >}}
 
--- a/spec-0000/index.md
+++ b/spec-0000/index.md
@@ -70,6 +70,15 @@ The situation has since improved due to improved installations via binary wheels
 
 ### Drop Schedule
 
+Below is an auto generated schedule, with recommended date for considering
+stopping support. We suggest that the next release in a given quarter is
+considered as the one removing support for a given item.
+
+In particular for Python releases, as all releases are around October, you may
+want to delay removal of Supports of an older Python version until your package
+work on the newly released Python version, thus keeping the number of minor
+version of python supported the same for all releases.
+
 {{< include-md "schedule.md" >}}
 
 ## Notes
--- a/spec-0000/index.md
+++ b/spec-0000/index.md
@@ -70,12 +70,11 @@ The situation has since improved due to improved installations via binary wheels
 
 ### Drop Schedule
 
-Below is an auto generated schedule, with recommended date for considering
-stopping support. We suggest that the next release in a given quarter is
+Below is an auto generated schedule with recommended dates for dropping support. We suggest that the next release in a given quarter is
 considered as the one removing support for a given item.
 
 In particular for Python releases, as all releases are around October, you may
-want to delay removal of Supports of an older Python version until your package
+want to delay removal of support of an older Python version until your package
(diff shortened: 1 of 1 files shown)


jarrodmillman (approved):

spec-0000/index.md
```diff
@@ -70,6 +70,15 @@ The situation has since improved due to improved installations via binary wheels
 
 ### Drop Schedule
 
+Below is an auto generated schedule, with recommended date for considering
+stopping support. We suggest that the next release in a given quarter is
```
lagru, line 74: The grammar might be off here? How about "..., with recommended dates at which dropping support should be considered." Not an expert either tough.

Carreau:
yeah i'm not a native english either, so better phrasing welcome.

spec-0000/index.md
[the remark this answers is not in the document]
jarrodmillman: A recommendation is just a suggestion. I am not sure our recommendation is that folks consider the "recommendation." I would just make the recommendation with the understanding that a recommendation is something to consider.

The more qualifying language we add the more legal sounding these documents become and the less clear the guidance is. We can add some, but I just don't want it to go to far.

spec-0000/index.md
```diff
@@ -70,6 +70,15 @@ The situation has since improved due to improved installations via binary wheels
...
+considered as the one removing support for a given item.
+
+In particular for Python releases, as all releases are around October, you may
+want to delay removal of Supports of an older Python version until your package
```
jarrodmillman, line 78:
```suggestion
want to delay removal of support of an older Python version until your package
```

spec-0000/index.md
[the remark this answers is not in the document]
Carreau: The min problem I was starting to adresse is a discussion a couple of hours ago in Napari, where the feeling were both: 

1) We should drop support on October 5th for Python 3.9 (and I don't think we want to convey an exact date).
2) Some people I felt were shocked to have such an early stop-support date, when it was not clear yet when Python 3.12 would be out (it is now, but the whole ecosystem needs rebuild.)

munkm:
A more broad comment on the wording we're converging on here: is it really the technicalities of when to drop that is confusing? In my experience the big issue is what people think of when "drop" is brought up. Maybe we want to add a disclaimer that says something to the effect of: 

> Disclaimer on this SPEC: Drop here is not meant to convey removing support for specific versions of upstream packages. Instead, we intend to ease maintenance burden by providing a schedule where upstream version support is not guaranteed to be supported. This way maintainers have a specific window of support they may commit to without adding maintenance burden of deprecations, etc.

spec-0000/index.md
```diff
@@ -70,6 +70,14 @@ The situation has since improved due to improved installations via binary wheels
 
 ### Drop Schedule
 
+Below is an auto generated schedule with recommended dates for dropping support. We suggest that the next release in a given quarter is
+considered as the one removing support for a given item.
+
+In particular for Python releases, as all releases are around October, you may
+want to delay removal of support of an older Python version until your package
+work on the newly released Python version, thus keeping the number of minor
+version of python supported the same for all releases.
```
bsipocz, line 79: I would be even more conservative here, and would say that it has to fully work, aka optional dependencies should be compatible, too.
  
```suggestion
"you may want to delay the removal of support of an older Python version until your package fully works on the newly released Python, thus keeping the number of supported minor versions of Python the same for your package."
```

bsipocz (approved):
Overall I like the softening as this drop schedule is often misquoted and used to somewhat aggressively demand a drop (which is quite annoying for developer infrastructure or when a lot of the optional dependencies are not yet compatible with the new version).

Carreau:
I think that would make things more confusing.
in practice it is removing support. I mean if you don't test on CI it might
as well be broken.

For me in IPython, when I drop, i'll bump flags in ou upgrades, and that
will likely remove conditionals like if version_info < ... , so the package
won't work.

let say it work and you put python_requires>= , then even if supported it
won't install.

I think we can leave "drop" to interpretation.

IMHO what maintainer may not realise is that often dropping from main mean
it will get drop from stable a few month later once there is a new release.

bsipocz:
I agree with @Carreau here. While some parts (maybe even 99.9%) keep working when dropping support, removing from CI but not explicitly changing python_requires tended to lead to more problems and a lot of headaches of sorting out bugreports. 

I also like to be very explicit with minimum versions for dependencies, even if an older one might work, I bake in the minimum that I test with and don't rely on wildcarded versions either.

spec-0000/index.md
```diff
@@ -70,6 +70,11 @@ The situation has since improved due to improved installations via binary wheels
 
 ### Drop Schedule
 
+Below is an auto generated schedule with recommended dates for dropping support. We suggest that the next release in a given quarter is
+considered as the one removing support for a given item.
+
+"you may want to delay the removal of support of an older Python version until your package fully works on the newly released Python, thus keeping the number of supported minor versions of Python the same for your package."
```
Carreau, line 76:
```suggestion
You may want to delay the removal of support of an older Python version until your package fully works on the newly released Python, thus keeping the number of supported minor versions of Python the same for your package.
```

Just removing the quotes.

Carreau: It's text, text is always in quotes....

munkm:
> While some parts (maybe even 99.9%) keep working when dropping support, removing from CI but not explicitly changing python_requires tended to lead to more problems and a lot of headaches of sorting out bugreports.

That's definitely not what I was trying to imply here. I was trying to make it clear that this SPEC isn't intended to add maintenance burden for lower support versions (in both ways -- by not keeping some lower version supported for all time, but also not that dropping support needs to be some huge deal every quarter)? That was my impression from our conversations. Is there a way to be clearer about this so it isn't interpreted aggressively like we've seen in some cases?

bsipocz:
Ahh, ok, sorry, then we're all on the very same page :)

PR #273, opened 2023-10-02
SPEC 0: Soften explicit drop schedule.

Carreau:
This does only change the presenting of the drop information:

 - reduplicate when the same package was released twice in the same quarter.
 - Remove the day of month. I don't think it helps conveying the flexibility that the spec

-- 

This also regenerate the files


jarrodmillman:
The format of `spec-0000/schedule.md` needs to be fixed. It doesn't pass the linting and I suspect the rendered text isn't what you intended:
2023-10-02T10:05:37,451492976-07:00

You may want to use the `yamltotable` shortcode:
https://theme.scientific-python.org/shortcodes/#yamltotable

Carreau:
Yeah, I'm trying 4 spaces to have pre blocks, it's not the best. 

Screenshot 2023-10-02 at 19 17 44

I'm going to see if it supports tables.

munkm:
> Remove the day of month. I don't think it helps conveying the flexibility that the spec

I think this is a good idea and I'm in support of this change. 

Something I've been thinking about with respect to the flexibility of this SPEC is the implications of the word "drop". I wonder if we add language saying that "drop" doesn't mean explicitly removing support, but rather not continuing support for that versioning. What do you think?

Carreau:
Better as a table:

Screenshot 2023-10-02 at 19 38 53

spec-0000/schedule.md
```diff
@@ -1,98 +1,115 @@
...
-- 24 Jun 2025: drop scipy 1.11.0 (initially released on Jun 25, 2023)
-
-2025 – Quarter 4:
-
-- 23 Oct 2025: drop python 3.11 (initially released on Oct 24, 2022)
-
-2026 – Quarter 4:
-
-- 01 Oct 2026: drop python 3.12 (initially released on Oct 02, 2023)
+#### 2023 - Quarter 4:
```
bsipocz, line 1: I feel that the previous quarter could be kept around for a bit of an overlap.

spec-0000/schedule.md
```diff
@@ -1,98 +1,115 @@
...
+
+Recommend drop support for:
```
bsipocz, line 3: The rendered version could maybe benefit from a tiny bit of different formatting, either leaving more space between this line and the table itself, or making this line with a different typeface (italic should be enough, no need to make it a heading).

bsipocz (approved):
I really like the cleaner look of your table solution than what we currently have.

spec-0000/schedule.md
Carreau: Yeah, spacing need css change, as prettify reformat the markdown anyway.

spec-0000/SPEC0_versions.py
```diff
@@ -124,31 +124,45 @@ def get_release_dates(package, support_time=plus24):
...
-            current_quarter = qt
 
-        fh.write(
-            f"- {dates[1].strftime('%d %b %Y')}: drop {package} (initially released on {dates[0].strftime('%b %d, %Y')})\n"
-        )
+print("Saving drop schedule to schedule.md")
+with open("schedule.md", "w") as fh:
+    for quarter in sorted(set(dq.index.get_level_values(0))):
+        fh.write("#### " + str(quarter).replace("Q", " - Quarter ") + ":\n\n")
+        fh.write("Recommend drop support for:\n\n")
```
bsipocz, line 150: Maybe then just make it a heading? 
```suggestion
        fh.write("###### Recommend drop support for:\n\n")
```

spec-0000/schedule.md
bsipocz: what about https://github.com/scientific-python/specs/pull/273#discussion_r1344398340 then?

jarrodmillman: You can put the markdown inside of
```
<!-- prettier-ignore-start -->
```
and
```
<!-- prettier-ignore-end -->
```
to prevent prettier from messing with your formatting.

Feel free to let @alphapapa know if you would like any css updates to the theme. You are also welcome to just make a PR if you prefer.

spec-0000/SPEC0_versions.py
Carreau: Let's try

Carreau:
I've tried reading 1 quarter in the past.

bsipocz:
I quite like what I see under the current preview link, not sure why prettier doesn't like it though.

stefanv:
Easiest is to run the generated result through prettier and look at the resulting diff.

jarrodmillman:
Or ask the pre-commit bot to autofix it.

stefanv:
If it's for a generated file, you'd have to run that each time you produce the file.

jarrodmillman:
Here is what prettier does. For example, it changes this
```
|    |    |    |
|----|----|----|
|ipython        |7.26.0 to 7.28.0   |released Aug 2021 and Sep 2021|
|networkx       |2.6                |released Jul 2021|
|pandas         |1.3.0              |released Jul 2021|
|scikit-learn   |1.0                |released Sep 2021|
|xarray         |0.19.0             |released Jul 2021|
```
to
```
|              |                  |                                |
| ------------ | ---------------- | ------------------------------ |
| ipython      | 7.26.0 to 7.28.0 | released Aug 2021 and Sep 2021 |
| networkx     | 2.6              | released Jul 2021              |
| pandas       | 1.3.0            | released Jul 2021              |
| scikit-learn | 1.0              | released Sep 2021              |
| xarray       | 0.19.0           | released Jul 2021              |
```

jarrodmillman:
There is a blank line at the end of `schedule.md` that I still need to fix. After that, this should be ready to go.

jarrodmillman:
@Carreau @bsipocz I removed the final blank line in 201e92372b046ebb1062b98e527b47cb8e0816c8. The CI is passing now and the tables are formatted in the new style. I am sure the code can be improved. Feel free to fix my commits or just delete them and fix the PR in a different way.

jarrodmillman (approved):

bsipocz (approved):
The rendering looks good to me so I keep my approval for merging as I don't have strong opinions on how the code archives that
(I have the feeling that the astropy fixed width ascii table writer could recreate something like these in much fewer lines, but don't think it would be worth the effort of rewriting everything)

Issue #280, opened 2023-10-16
Revise wording to better convey the purpose of SPECs


cbrnr:
Following up the discussion in https://github.com/scientific-python/lazy_loader/pull/70, the way that SPECs are presented on the Scientific Python website can be misleading. It puts some (a lot of?) pressure on packages to adopt all SPECs, at least that's my impression. Since this is not intended, I think some modifications to the website might help clarify this. Here are some examples:

- The introductory sentence on the [main SPEC page](https://scientific-python.org/specs/) says: "SPECs provide operational guidelines for projects in the scientific Python ecosystem." If I develop a project in the scientific Python ecosystem, this sounds like I need to follow these guidelines. I guess the term "guidelines" is too strong, maybe "recommendations" or "suggestions"? I think it would be great if the first sentence already clarified that it's OK to not adopt any SPECs if it is not feasible, and that a project would still be part of the scientific Python ecosystem.
- The list of recommendations lists all SPECs as well as projects which have endorsed them. I think this is exactly the thing that's misleading projects into thinking that every good projects should strive to adopt as many SPECs as possible. After all, they get featured prominently on the website then. Honestly, I would remove all endorsing projects here, and maybe create a separate document, which describes projects that have adopted certain specs. This would give more room for presenting the pros and cons of each SPEC, and maybe make it clear that it's OK if a projects opts to not adopt a certain SPEC. It would also make it harder to see the amount of SPECs each project has adopted. Or maybe list projects which have supported a specific SPEC at the bottom of each SPEC page.
- On each SPEC page, and I'm now mainly talking about SPEC 1 (lazy loading), tone down the wording so that it is clear that no project has to adopt that SPEC. This also includes listing the pros and cons, which are already discussed, but maybe not prominently enough (only the advantages are listed near the top, the disadvantages get a short Caveats section much further into the text. For example, the introduction says: "This SPEC recommends a lazy loading mechanism—targeted at libraries—that avoids import slowdowns and provides explicit submodule exports." The libraries part is a bit unclear – when is a package a library? So after reading this, the message I get is that lazy loading is recommended. Period. I'd also include more caveats in this section (instead of footnotes), such as the rejection of the lazy import PEP, comments by some Python core devs why they are probably never going to include the mechanism, etc. This is extremely important information, which I had completely missed when we were discussing including lazy loading in MNE. At that time I just skimmed the SPECs page and thought "yeah, this sounds great, who doesn't want reduced import times for free?".

[...]

PR #283, opened 2023-10-18
Add caveats about adopting SPEC 1

jarrodmillman:
See #280. @cbrnr Would you mind suggesting a sentence or two (more if necessary) providing some concerns projects should consider before adopting this SPEC?

### DIFF
--- a/spec-0001/index.md
+++ b/spec-0001/index.md
@@ -50,7 +50,7 @@ Endorsing this SPEC means agreeing, in principle, with the advantages of lazy lo
 ### Ecosystem Adoption
 
 {{< notice warning >}}
-Not all projects should use lazy loading.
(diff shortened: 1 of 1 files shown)


spec-0001/index.md
```diff
@@ -49,6 +49,10 @@ Endorsing this SPEC means agreeing, in principle, with the advantages of lazy lo
 
 ### Ecosystem Adoption
 
+{{< notice warning >}}
+Not all projects should use lazy loading.
```
stefanv, line 53:
```suggestion
We do not recommend lazy loading for all projects. E.g., small projects, or projects with low import overheads do not need it. Lazy loading is useful when you are concerned about subpackage import times, but also want to make those subpackages available for interactive exploration in, e.g., IPython.
```

stefanv (approved):

PR #291, opened 2023-12-21
SPEC-4 endorsement for scikit-learn

jjerphan:
scikit-learn has adopted SPEC 4 but has not yet endorsed it.

This is currently being discussed on the scikit-learn's internal mailing list, and I would wait for several maintainers' approval and no veto before merging this PR.

### DIFF
--- a/spec-0004/index.md
+++ b/spec-0004/index.md
@@ -18,6 +18,7 @@ endorsed-by:
   - scikit-image
   - scipy
   - xarray
+  - scikit-learn
 ---
 
 ## Description


GaelVaroquaux (approved):
LGTM, as a scikit-learn core dev, I approve this endorsement.

jeremiedbb (approved):

thomasjpfan (approved):

betatim (approved):

tupui (approved):
Thank you for the endorsement and to all for stopping by to confirm 🚀

jjerphan:
Thank you all for your responsiveness!

I think the maintainers of scikit-learn need to discuss and see whether the other SPECs of Scientific Python can be implemented before endorsing them.

GaelVaroquaux:
On Fri, Dec 22, 2023 at 01:27:17AM -0800, Julien Jerphanion wrote:
> I think the maintainers of scikit-learn need to discuss and see whether the other SPECs of Scientific Python can be implemented before endorsing them.

What's the value for scikit-learn? How does it make our job easier or more effective?

jjerphan:
Generally, the motivation is given in the "motivation" section of each SPEC.

For instance, here is the one of [SPEC 0](https://scientific-python.org/specs/spec-0000/#motivation):

> Limiting the scope of supported dependencies is an effective way for packages to limit maintenance burden. Combinations of packages need to be tested, which impacts also on continuous integration times and infrastructure upkeep. Code itself also becomes more complicated when it has to be aware of various combinations of configurations.
>
> Adoption of this SPEC will ensure a consistent support policy across packages, and reduce the need for individual projects to divise similar policies.
>
> Ultimately, reduced maintenance burden frees up developer time, which translates into more features, bugfixes, and optimizations for users.

Probably not all SPECs are relevant for scikit-learn (e.g. [SPEC 1](https://scientific-python.org/specs/spec-0001/) was judged not to be relevant in https://github.com/scikit-learn/scikit-learn/pull/26440), but some of them might be of value for scikit-learn, and having a base project like scikit-learn endorse some SPECs re-enforces the entire ecosystem.

What do you think?

stefanv:
Re: implementing vs endorsing, the first does not need to preceed the second. 

The SPECs are meant to signal best practices. If your project thinks a SPEC describes how something should be done, whether you do it that way yourself or not, you can endorse it.

The goal is to reach some community consensus around best practices, making it easier to grow the ecosystem, and making it easier for maintainers to know what to do (since there's a central place that documents such practices).

PR #293, opened 2024-01-19
Update pre-commit repos

jarrodmillman:
Switch to quarterly updates and manually updated via:
```
import yaml
import subprocess

file = ".pre-commit-config.yaml"

with open(file) as f:
    data = yaml.safe_load(f)

repos = []
for repo in data["repos"]:
    if "mirrors-prettier" not in repo["repo"]:
        repos.append("--repo")
        repos.append(repo["repo"])

command = ["pre-commit", "autoupdate", "--freeze"] + repos
print(" ".join(command))
subprocess.call(command)
```

Work around to avoid updating to an alpha version of https://github.com/pre-commit/mirrors-prettier.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,7 +17,7 @@ repos:
       - id: check-added-large-files
 
   - repo: https://github.com/psf/black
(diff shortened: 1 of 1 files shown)


jarrodmillman:
See https://github.com/scientific-python/scientific-python.org/pull/549.

bsipocz (approved):

PR #298, opened 2024-02-06
Add Lars and Pamphile as SPEC-6 co-authors

stefanv:
I forgot to update the authors list for SPEC6. Lars and Pamphile gave significant input.

@lagru @tupui let me know if this is OK?


tupui (approved):
Thank you 😃 Good for me 👍

lagru (approved):
Thanks for the consideration, I don't feel like I added that much but I'm okay with this of course. :)

jarrodmillman (approved):

PR #306, opened 2024-03-06
Add numbers for sorting spec list

jarrodmillman:


bsipocz (approved):

stefanv:
If we're going to add numbers, shall we just remove it from the title and render that in programmatically?

stefanv:
Also, maybe call that parameter `spec`, then the header becomes:

```
---
spec: 0
title: Minimum Supported Dependencies
...
```

PR #307, opened 2024-03-06
Delete old content and replace with basic skeleton

### LINKED ISSUE #229: Handle SPEC 2
- [x] Remove attached website @jarrodmillman 
  - [x] #239 @jarrodmillman 
  - [x] Save meeting notes @jarrodmillman archived https://github.com/scientific-python/api-dispatch.scientific-python.org
  - [x] remove website from netlify @jarrodmillman 
- [ ] Update SPEC text @jarrodmillman #307

jarrodmillman:
Close #229.

### DIFF
--- a/spec-0002/index.md
+++ b/spec-0002/index.md
@@ -3,11 +3,8 @@ title: "SPEC 2 — API Dispatch"
 number: 2
 date: 2021-12-16
 author:
-  - "Ivan Yashchuk <ivan.yashchuk@quansight.com>"
-  - "Ralf Gommers <rgommers@quansight.com>"
   - "Jarrod Millman <millman@berkeley.edu>"
-  - "Stéfan van der Walt <stefanv@berkeley.edu>"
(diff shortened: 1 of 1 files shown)


spec-0002/index.md
```diff
@@ -3,50 +3,70 @@ title: "SPEC 2 — API Dispatch"
...
+
+<!--
+Briefly discuss what it means for a core project to endorse this SPEC.
+-->
+
+### Ecosystem Adoption
+
+<!--
+Briefly discuss what it means for a project to adopt this SPEC.
+-->
```
tupui, line 35: Non blocking comment: we were discussing that maybe these don't have to be in the skeleton. Endorsement and adoption is defined globally for all SPECs and should mostly be the same. Hence, unless there is a distinction specific to this SPEC which needs to be made, I would remove these sections.

tupui (approved):
LGTM 👍

spec-0002/index.md
stefanv: @tupui Do you mind filing this as an issue for discussion?

PR #318, opened 2024-05-15
SPEC0 - Adjust past quarter to 9 month and regenerate

Carreau:
Only 6 month is a bit short, like right now we do not include what should have been dropped in last quarter of 2023 even if we are less than 6 month.

We also use pandas to always use a cutoff date that is a quarter to avoid content of a section to change depending on when this is regenerated.


tupui (approved):
+1 from me

bsipocz (approved):
Looks totally reasonable to me. Thanks!

stefanv (approved):

PR #319, opened 2024-06-04
Add Matplotlib endorsements

QuLogic:
We endorse SPEC 0 (minimum dependency versions) and SPEC 4 (nightly wheels), at the least.

These are things we've already implemented, so I hope I don't need to provide any extra confirmation.


tupui (approved):
yay 🎉

PR #326, opened 2024-06-07
SPEC 12: Formatting mathematical expressions

tupui:
This follows the proposal on the forum https://discuss.scientific-python.org/t/spec-12-formatting-mathematical-expressions

See other linked discussions as well.


tupui:
cc @mdhaber @stefanv @jarrodmillman and add yourself as co-authors of course 😉

spec-0012/index.md
```diff
@@ -0,0 +1,125 @@
...
+
+- There a space before and after `-` and `+`. Except if
+  the operator is used to define the sign of the number;
+
+  ```
+  a + b
+  -a
+  ```
+
+- Within a group, if operators with different priorities are used, add whitespace around the operators with the lowest priority(ies).
```
mdhaber, line 42: According to the definition of "group", it is not possible for operators to have different priority.

I believe the definition of "group" was supposed to be something more like a sequence of operations that relies on implicit order of operations rules. Examples include:
- a logical line
- operations within parentheses
- the expression and for/if clauses of a list comprehension

spec-0012/index.md
```diff
@@ -0,0 +1,125 @@
...
+## Implementation
+
+The following rules must be followed.
+These rules respect and complement the PEP8 (relevant sections includes
+[id20](https://www.python.org/dev/peps/pep-0008/#id20) and
+[id20](https://www.python.org/dev/peps/pep-0008/#id28)).
+
+We define a _group_ as a collection of operators having the same priority.
+e.g. `a + b + c` is a single group, `a + b * c` is composed of two groups `a`
+and `b * c`. A group is also a collection delimited with parenthesis.
```
mdhaber, line 30: See https://github.com/scientific-python/specs/pull/326/files#r1631759561

spec-0012/index.md
```diff
@@ -0,0 +1,125 @@
...
+  ```
+
+- When splitting an equation, new lines should start with the operator linking
+  the previous and next logical block. Single digit on a line are forbidden.
+
+  ```
+  (
+      a/b
+      + c*d
+  )
```
mdhaber, line 80: I think this should show an example where splitting the line makes sense. Consider referring to PEP8 "[Should a Line Break Before or After a Binary Operator](https://peps.python.org/pep-0008/#should-a-line-break-before-or-after-a-binary-operator)".

I'm not sure if the use of the term "logical block" is correct. This is a single logical line split across multiple physical lines.

spec-0012/index.md
```diff
@@ -0,0 +1,125 @@
...
+  (a*b) * (c*d)
+  ```
+
+- Operators within a group are ordered from the lowest to the highest priority.
+  If this is technically an issue (e.g. restriction on the AST), add
+  parenthesis or spaces.
+
+  ```
+  a/d*b**c
+  a*(b**c)/d
```
mdhaber, line 68: This doesn't appear to follow the rule.

spec-0012/index.md
```diff
@@ -0,0 +1,125 @@
...
+ element.
+
+  ```
+  a*b
+  (a*b) * (c*d)
+  ```
+
+- Operators within a group are ordered from the lowest to the highest priority.
+  If this is technically an issue (e.g. restriction on the AST), add
+  parenthesis or spaces.
```
mdhaber, line 64: I think the reason for the exception - if there needs to be one - should be more explicit. 

This rule conflicts with the previous rule about space around operators `*` and `/`.

spec-0012/index.md
```diff
@@ -0,0 +1,125 @@
...
+
+- There is no space before and after operators `*` and `/`. Only exception is if the expression consist of a single operator linking two groups with more than one
+ element.
+
+  ```
+  a*b
+  (a*b) * (c*d)
+  ```
+
+- Operators within a group are ordered from the lowest to the highest priority.
```
mdhaber, line 62: I'm not sure "group" here is being used in the same sense as elsewhere. I remember discussing in person the idea that `a/d*b**c` is preferable to `a*b**c/d` unless there are explicit parentheses, like `(a*b**c)/d`, but it would not be wrong to do `a/d*b**c + e/f*g**h`, yet the plus comes after higher priority operators.

spec-0012/index.md
```diff
@@ -0,0 +1,125 @@
...
+  ```
+
+- There is no space before and after `**`.
+
+  ```
+  a**b
+  ```
+
+- There is no space before and after operators `*` and `/`. Only exception is if the expression consist of a single operator linking two groups with more than one
+ element.
```
mdhaber, line 55: There seems to be another exception below.
https://github.com/scientific-python/specs/pull/326/files#r1631763770

According to "Only exception is if the expression consist of a single operator linking two groups", there is no exception for:
```
(a*b)*(c*d)*(e*f)
```
because there are three groups? Or do you mean that you need a space when any binary operator is linking two explicit "groups" enclosed by parentheses?

spec-0012/index.md
```diff
@@ -0,0 +1,125 @@
...
+    *np.prod(
+        0.5*(2. + abs(z_ij[i1, :]) + abs(z_ij) - abs(z_ij[i1, :] - z_ij)), axis=1
+    )
+)
+```
+
+```python
+# bad
+i = i + 1
+submitted += 1
```
mdhaber, line 106: Why are these bad?

spec-0012/index.md
```diff
@@ -0,0 +1,125 @@
...
+    * np.prod(
+        0.5 * (2.0 + abs(z_ij[i1, :]) + abs(z_ij) - abs(z_ij[i1, :] - z_ij)), axis=1
+    )
+)
+```
+
+## Notes
+
+These formatting rules do not make any consideration in terms of performances
+nor precision. The scope is limited to styling.
```
mdhaber, line 125: Do you mean that these rules only apply when there are no precision/performance reasons to break them, and exceptions should be declared if needed to achieve precision or performance goals?

spec-0012/index.md
tupui: In your example it means no spaces because we have 3 groups.

And if we have 2 groups then each group must have at least one operator in it (ie not just a variable or single number).

spec-0012/index.md
tupui: How so?

spec-0012/index.md
tupui: A general thought, rules could conflict if we ask them to be applied one after the order in a strict order.

spec-0012/index.md
tupui: They are good, for now o just copy pasted the whole block in Black. But yes we should remove the correct ones from there.

spec-0012/index.md
tupui: See my other comment.

tupui (commented):
Thanks Matt. If you are going to be an author, feel free to just go ahead and commit changes 😉

tupui:
Hi @charliermarsh @ambv 👋 I had pinged you both on Twitter some years back about a standard for mathematical equations. At the time you both said this could interest you for both Ruff and Black if we, the Scientific Python community, could come to an agreement.

To be clear, I am of course in no position to ask for any commitment from you and I just hope that you find the topic interesting enough.

This is getting into shape and we have a preliminary draft we would like to present you 😃

Any comments would greatly help and in the end this can also only work if both Ruff and Black would be able and willing to implement this specification 🙏 

Thanks again both!

charliermarsh:
Awesome, I look forward to reading + engaging here.

mdhaber:
Thanks @charliermarsh. I think the main question right now is whether this is close to being precise enough to be implementable. For example, I don't have much background with ASTs, so perhaps you can suggest ideas that would replace the notions like "implicit subexpressions" I attempt to define.

spec-0012/index.md
```diff
@@ -0,0 +1,185 @@
...
+0. Unless otherwise specified, rely on the implicit order of operations;
+   i.e., do not add extraneous parentheses. For example, prefer `u**v + y**z`
+   over `(u**v) + (y**z)`, and prefer `x + y + z` over `(x + y) + z`. A full
+   list of implicit operator priority levels is given by
+   [Operator Precedence](https://docs.python.org/3/reference/expressions.html#operator-precedence)
+1. Always use the `**` operator and unary `+`, `-`, and `~` operators *without*
+   surrounding whitespace. For example, prefer `-x**4` over `- (x ** 4)`. 
+2. Always surround non-PEMDAS operators with whitespace, and always make the priority of
+   non-PEMDAS operators explicit. For example, prefer `(x == y) or (w == t)` over
+   `x==y or w==t`.[^1]
```
charliermarsh, line 95: Does this imply that we _should_ add parentheses here (as in: is this an exception to Rule 0)?

spec-0012/index.md
```diff
@@ -0,0 +1,185 @@
...
+to be "PEMDAS operators", and we will also include the unary `+`, `-`, and `~` in
+this category for convenience. The order of operations of PEMDAS operators is typically
+taught in primary school and reinforced throughout a programmer's training and
+experience, so it is assumed that most programmers are comfortable relying on the
+implicit order of operations of expressions involving a few PEMDAS operations. Implicit
+order of operations becomes less obvious as the number of distinct operator priority
+levels increases and when multiple non-PEMDAS operators are involved. Portions of this
+acronym, namely MD and AS, will be used below to refer to the corresponding operators.
+
+## Implementation
```
charliermarsh, line 76: Should any of these rules differ based on the expression type? E.g., all of the examples below use `Name` nodes (like `x`, `y`, etc.). What if the expression uses calls or subscripts or similar? Like `f() ** 2`?

spec-0012/index.md
```diff
@@ -0,0 +1,185 @@
...
+```python3
+for j in range(1, i + 1)
+```
+The output expression `j` is one explicit expression and the input sequence
+`range(1, i + 1)` is another. 
+
+A "subexpression" is subset of an expression that is either explicit or could
+be made explicit (i.e. with parentheses) without affecting the order of
+operations. In the example above, `j` and `range(1, i + 1)` can also be
+referred to as explicit subexpressions of  the whole expression, and `1` and
```
charliermarsh, line 39: "...of  the whole expression"

Assuming I'm reading this correctly, I would be careful here as `for j in range(1, i + 1)` is not an "expression" in the sense of the Python AST. `for` is a kind of "statement".

spec-0012/index.md
```diff
@@ -0,0 +1,185 @@
...
+`i + 1` are explicit subexpressions of the expression `range(1, i + 1)`. `i` and
+`1` are "implicit" subexpressions of `i + 1`: they could be written as explicit
+subexpressions `(i)` and `(1)` without affecting the order  of operations, but they
+are not explicit as written.
+
+As another example, in `x + y*z`, `y*z` is a subexpression because it could be made
+explicit as in `x + (y*z)` without changing the order of operations. However, `x + y`
+would not be a subexpression because `(x + y)*z` would change the order of operations.
+Note that `x + y*z` as a whole may also be referred to as a "subexpression" rather than
+an "expression" even though `(x + y*z)` is not a proper subset of the whole.
```
charliermarsh, line 49: What does "proper subset of the whole" mean? Sorry!

spec-0012/index.md
```diff
@@ -0,0 +1,185 @@
...
+   surrounding whitespace. For example, prefer `-x**4` over `- (x ** 4)`. 
+2. Always surround non-PEMDAS operators with whitespace, and always make the priority of
+   non-PEMDAS operators explicit. For example, prefer `(x == y) or (w == t)` over
+   `x==y or w==t`.[^1]
+3. Always surround AS operators with whitespace.
+4. Typically, surround MD operators with whitespace, except in the following situations.
+   - When there are lower-priority operators (namely AS) within the same compound
+     expression. For example, prefer `z = -x * y**t` over `z = -x*y**t`, but
+     prefer `z = w + x*y**t` over `z = w + x * y**t` due to the presence of the
+     lower-priority addition operator.
```
charliermarsh, line 101: That's neat. So this would be correct: `a = b + c*d**e + f*g**h`?

charliermarsh: What about `a = b + (c * d**e)*(f * g**h)`?

charliermarsh: (I know you're not looking for critiques right now, I'm just making sure I understand the rules.)

spec-0012/index.md
```diff
@@ -0,0 +1,185 @@
...
+   - The division operation would be written mathematically as a fraction with a
+     horizontal bar. For example, prefer `z = t/v * x/y` over `z = t / v * x / y`
+     if this would  be written mathematically as the product of two fractions,
+     e.g. $\frac{t}{v} \cdot \frac{x}{y}.
```
charliermarsh, line 105: I think this would need a more precise definition to implement.

spec-0012/index.md
```diff
@@ -0,0 +1,185 @@
...
+      performed first for numerical reasons or because `(x + y)` is a conceptually
+      important quantity. In such cases, consider adding a comment, e.g.
+      ```python3
+      t = (x + y) - z  # perform `x + y` first for precision
+      ```
+      or breaking the expressions into separate logical lines, e.g.
+      ```python3
+      w = x + y
+      t = w - z
+      ```
```
charliermarsh, line 178: Just to be explicit: is the intention that autoformatters (like Ruff) would support automatically formatting per this spec?

charliermarsh:
My initial reaction is that the definitions are sufficiently clear (minor some comments inline) to be implementable. I don't know that it would be _easy_ but I don't see ambiguities.

By the way, if you want to look at the AST for any of these, we have a playground: https://play.ruff.rs/85bce11a-11c4-4b66-b7ef-bf13cfb2d518. We deviate from the CPython AST in some places, but for binary expressions it should be roughly the same IIRC.

MichaReiser:
I only skimmed through it, but it's great to see more people think about improving expression formatting. I don't recall Ruff's and Black's rules explicitly. Do you know where this style guide deviates from Black/Ruff (evaluating one should be enough)? 

I'm especially interested in the formatting of very long expressions. I've written up some of my thoughts in https://github.com/psf/black/issues/4123

spec-0012/index.md
MichaReiser: I think this would be hard or even impossible because the formatter has no domain knowledge and requires explicit rules on how to deal with conflicts

spec-0012/index.md
```diff
@@ -0,0 +1,185 @@
...
+The output expression `j` is one explicit expression and the input sequence
+`range(1, i + 1)` is another. 
+
+A "subexpression" is subset of an expression that is either explicit or could
+be made explicit (i.e. with parentheses) without affecting the order of
+operations. In the example above, `j` and `range(1, i + 1)` can also be
+referred to as explicit subexpressions of  the whole expression, and `1` and
+`i + 1` are explicit subexpressions of the expression `range(1, i + 1)`. `i` and
+`1` are "implicit" subexpressions of `i + 1`: they could be written as explicit
+subexpressions `(i)` and `(1)` without affecting the order  of operations, but they
```
mdhaber, line 42:
```suggestion
subexpressions `(i)` and `(1)` without affecting the order of operations, but they
```

spec-0012/index.md
```diff
@@ -0,0 +1,185 @@
...
+A "compound" expression is an expression involving more than one operator
+priority level without considering the contents of explicit subexpressions.
+For example, 
+- `x + y - z` is a simple expression because `+` and `-` have the
+same priority level. There are no explicit subexpressions to be ignored.
+- `x * (y + z)` is also a simple expression because there is only one operator
+between `x` and the explicit subexpression `(y + z)`; we ignore the contents - and
+especially the operator - within the explicit subexpression; conceptually, it may
+regarded as `(...)`.
+- `x * y + z` is a compound expression; there are two operators and no explicit
```
mdhaber, line 62:
```suggestion
- `x*y + z` is a compound expression; there are two operators and no explicit
```

mdhaber (commented):
Thanks very much for your thoughts!

>  I don't recall Ruff's and Black's rules explicitly. Do you know where this style guide deviates from Black/Ruff (evaluating one should be enough)?

@MichaReiser I know it deviates from Black as-is. For instance, in Black
```
a = b + c*d**e + f*g**h
```
would become
```
a = b + c * d**e + f * g**h
```

This is actually one of the main motivations - Black seems to just add whitespace around everything and call it a day. Originally, it even added space around exponentiation (gh-538). This guide suggests that sometimes omitting whitespace or adding parentheses is important for readability.

I don't think it goes against any core Ruff rules because it doesn't conflict with PEP8, but please correct me @charliermarsh.

spec-0012/index.md
charliermarsh: Thanks, this makes sense! I think it will be clear to those that are familiar with the term.

spec-0012/index.md
mdhaber: Here's another possibility w.r.t. multiplication and division: use whitespace except when there are lower-priority operations in the compound subexpression; in that case, either omit the whitespace or add parentheses. I can see cases in which either might be preferred, so it might be best for the rule to just require that one of those two techniques be used to visually reinforce the operator priority.

MichaReiser:
> I don't think it goes against any core Ruff rules because it doesn't conflict with PEP8, but please correct me @charliermarsh.

Ruff implements a few E2XX and E3XX rules that we would have to double-check. E.g. [`missing-whitespace-around-operator`](https://docs.astral.sh/ruff/rules/missing-whitespace-around-operator/)

What's not entirely clear to me from the proposal is what your goal is:

* Is it an automatic formatter similar to Black that enforces consistent formatting? 
* Is it a set of lint rules that enforce a specific style for some expressions?

There are differences between the two and integrating your proposal into `ruff format` or `ruff check` imposes different challenges depending on what you're looking for. 

**Formatter**

The main goal of a formatter is to enforce consistent styling across a code base. The implication is that *letting the user decide on the formatting* generally isn't an option because it wouldn't lead to consistent formatting. The only exception to this in Ruff's (and Black's) formatter today is the use of magic-trailing-commas where the user can choose one style by adding a trailing comma. 

I think there are a few places where the proposal would have to become more opinionated if it should be implemented as an auto formatter.

The other challenge when it comes to building an auto-formatter is that the style guide must be consistent in its entirety. For example, ruff (and Black) place the opening parentheses on their own line:

```python
(
    t
    + (
        w
        + (x + (y + z))
    )
)
```

whereas this document proposes

```python
(t 
 + (w + (x + (y + z)))))
```
(I prefer your formatting, which is also what Prettier does, but let's ignore this for now). 

The main challenge with this formatting is that it is very different from how Black/Ruff places parentheses for all other syntaxes (list, parameters, arguments...). It makes binary expressions the "odd" ones, standing out from the rest of the code. I'm not concluding that this is a reason to rule out the style, but we would have to consider the style guide in the context of Ruff's entire style guide. 

**Linter**

The alternative is that the style guide is enforced by lint rules, each of which enforces a specific aspect. Some lint rules may have auto-fixes, but not all of them. 

From what I understand in the proposal, that's a better fit. The main integration challenge for Ruff is that lint rules enforcing a style different from the formatter create a compatibility problem: running `ruff check --fix` fixes the style in one direction, and `ruff format` undoes the changes. Style rules are also prone to being incompatible with other stylistic lint rules. E.g. are all proposed changes compatible with Ruff's `E2XX` and `E3XX` rules? 

This aren't reasons why Ruff can't support the proposed styles. But these are challenges that we would have to overcome and find solutions for before they can be added to ruff.


## Note
Another annoying thing that might be worth considering is if the specification wants to be explicit about where to place comments. How would expression be formatted if there's a leading comment? What about end of line comments?

mdhaber:
> Ruff implements a few E2XX and E3XX rules

Thanks. I'll take a look. But yes, this definitely conflicts with E225 (which actually seems to go against PEP8, TBH). 

> Style rules are also prone to being incompatible with other stylistic lint rules.

Rules can be turned on/off, though, right? After we determine which rules these conflict with, perhaps the SPEC would recommend which other rules would need to be disabled (e.g. E225) to adopt the SPEC?

> For example, ruff (and Black) place the opening parentheses on their own line

Hmm I intended for this to be explicitly agnostic about that specific point:
```
Although examples do not show the use of hanging indent, any of the indentation styles
allowed by [PEP 8 Indentation](https://peps.python.org/pep-0008/#indentation) are permitted by this SPEC.
```

As for where the line breaks go (Rules 7/8), I'm not sure they are explicit enough yet. So maybe we can leave those out initially, if that would help.

> ...where to place comments.

Hmm I didn't think about that. I don't have that in mind for the scope. Does this interact with the other rules besides those regarding line breaks? Maybe that's another reason to table the rules about line breaks.

> What's not entirely clear to me from the proposal is what your goal is:

The SPEC itself wouldn't say how the rules are to be enforced, but I think you are right that they would be better as part of a linter with some auto-fixes available.

---

So would that help to table rules 7 and 8 (about line break placement) and focus on enforcing the rest as linting rules (disabling other rules as needed)?

lucascolley:
> What's not entirely clear to me from the proposal is what your goal is:
> 
> Is it an automatic formatter similar to Black that enforces consistent formatting?
> Is it a set of lint rules that enforce a specific style for some expressions?

Speaking for myself, I would like to be able to enforce these rules via `ruff format`. Hopefully this along with https://github.com/astral-sh/ruff/discussions/8452 would mean we could start to consider an auto-formatter in projects like SciPy.

tupui:
Yep that's the idea, wish. Relying on Ruff's team to do it though as I am not sure any of us would have the know how to help on the implementation part.

MichaReiser:
> Rules can be turned on/off, though, right? After we determine which rules these conflict with, perhaps the SPEC would recommend which other rules would need to be disabled (e.g. E225) to adopt the SPEC?

They can, but everyone using `select=ALL` will see a warning unless they explicitly ignore the rule. That's why we tried to avoid adding any new rules that conflict with each other. We also want to categorize the rules long term and it's somewhat likely that we'll remove the `ALL` selector, so that conflicting rules are less of a problem, as long as they aren't enabled by default. The other solution has been to introduce a setting, e.g. similar to `pydocstyle`'s `convention` setting

 > As for where the line breaks go (Rules 7/8), I'm not sure they are explicit enough yet. So maybe we can leave those out initially, if that would help.

I would have to re-read them but it's okay to be vague if this proposal should be implemented as lint rules because Ruff can then just not enforce the style or ask users to use `noqa` to opt out. 

> Yep that's the idea, wish. Relying on Ruff's team to do it though as I am not sure any of us would have the know how to help on the implementation part.

Happy to help you to learn more about how the formatter works. 

Integrating this into the formatter certainly has a higher bar than into the linter. I would prefer to improve the existing style for the entire ecosystem if possible over having to maintain (and justify) two styles. If that's not possible, we then have to decide if Ruff should support multiple styles and what's the bar to justify why Ruff supports one style vs any other.

tupui:
Thanks @MichaReiser 🙇 Assuming we get this in (which we should if everyone is happy as to not be blocking anything), what do we do next? Would you be willing to take a shot at implementing these rules? First in the linter if I read correctly.

lucascolley:
> If possible, I would prefer to improve the existing style for the entire ecosystem over maintaining (and justifying) two styles. I think there's room to make expression formatting better (see psf/black#4123). If that's not possible, we then have to decide if Ruff should support multiple styles and what the bar is to justify why Ruff supports one style vs. any other.

IIRC, the `np.array` formatting discussion was what sparked the idea of a separate "scientific python" style. But if it could all be incorporated into the one style, great!

MichaReiser:
The very first step that I see is that you find agreement if you want this to be part of an auto-formatter (`ruff format`) or linter (`ruff check`) because both have different requirements on how specific a specification must be. Depending on that it is then our turn to figure out how this fits into ruff: e.g. we've been very intentional so far to not support multiple code styles in `ruff format`. 

> Would you be willing to take a shot at implementing these rules? First in the linter if I read correctly.

We're fairly thin-stretched right now with our work on UV, Ruff maintenance, and adding multifile analysis support to Ruff. We also have to solve many fundamental efforts like rule categorization first. Short answer: I'm very interested in improving code formatting for the scientific Python community, but it likely takes a while before I can make it a priority.

spec-0012/index.md
```diff
@@ -0,0 +1,185 @@
...
+     prefer `z = w + x*y**t` over `z = w + x * y**t` due to the presence of the
+     lower-priority addition operator.
+   - The division operation would be written mathematically as a fraction with a
+     horizontal bar. For example, prefer `z = t/v * x/y` over `z = t / v * x / y`
+     if this would  be written mathematically as the product of two fractions,
+     e.g. $\frac{t}{v} \cdot \frac{x}{y}.
+5. Considering the previous rules, only `**`, `*`, `/`, and the unary `+`, `-`, and `~`
+   operators can appear in implicit subexpressions without spaces. In such expressions,
+   - Use at most one unary operator, and if used, ensure that it is the leftmost operator.
+   - Use at most one `**` operator, and if used, ensure that it is the rightmost operator.
```
mdhaber, line 109:
```suggestion
   - Use at most one unary operator, and if used, ensure that it is the leftmost operator.
   - Use at most one `**` operator, and if used, ensure that it is the rightmost operator.
   - Use at most one `/` operator, and if used, ensure that it is the rightmost operator except for `**`.
```

lucascolley:
> The alternative is that the style guide is enforced by lint rules, each of which enforces a specific aspect. Some lint rules may have auto-fixes, but not all of them.
> 
> From what I understand in the proposal, that's a better fit. The main integration challenge for Ruff is that lint rules enforcing a style different from the formatter create a compatibility problem: running ruff check --fix fixes the style in one direction, and ruff format undoes the changes. Style rules are also prone to being incompatible with other stylistic lint rules. E.g. are all proposed changes compatible with Ruff's E2XX and E3XX rules?
> 
> This aren't reasons why Ruff can't support the proposed styles. But these are challenges that we would have to overcome and find solutions for before they can be added to ruff.

> The very first step that I see is that you find agreement if you want this to be part of an auto-formatter (ruff format) or linter (ruff check) because both have different requirements on how specific a specification must be.

@MichaReiser would the following be a feasible path forward here?

1. Implement lint rules which can detect everything this document seeks to disallow
	- this would cause conflict with some existing rules like https://docs.astral.sh/ruff/rules/missing-whitespace-around-operator/, which btw @mdhaber and I agree is not justified in claiming to be "according to PEP 8"[^1]
2. start working on auto-fixes where possible
3. if the coverage of auto-fixes is good enough to result in a consistent style, *potentially* integrate them into `ruff format`, either through a separate style or through the wider community agreeing to adopt these rules.
	- noting the difficulty above that this would create inconsistency with other code-blocks where mathematical expressions are not used

I understand that step 3 is a very high bar, and that step 1 is a lot of work in itself that is not necessarily going to happen any time soon, but is this a reasonable roadmap/wishlist? Or do you see problems that would make this path impossible?

[^1]: https://peps.python.org/pep-0008/#other-recommendations

MichaReiser:
I think a decision on whether to make this a set of lint rules or integrate it into the formatter is essential. 

* The formatter and linter work very differently. Implementing the style guide as lints rule first and then as part of the formatter means duplicated work.
* We would also have to keep maintaining the style guide lint rules because some users started to rely on them and use the flexibility that they can enable rules individually (unless this becomes a single massive rule)
* We don't have a good framework for style-guide like lint rules.

This is in addition to the challenges I mentioned before:

* We decided not to implement rules that conflict with the formatter because formatted code should never raise new lint violations. This property is essential to integrate the formatter into `check`. Some of the proposed style guide rules do conflict with the formatter. 
* The style guide conflicts with existing rules, and we can't just remove them. We have users who depend on them, even if the rules might go beyond what PEP-8 specifies. Arguably, we should never have added those rules,, but here we are.
* Implementing a formatting style guide is a multi-week, if not multi-month, project. It also increases the complexity to push any new formatting changes because they now need to fit into and be tested with multiple styles guides.
* I'm very hesitant about introducing a new style guide into the formatter because it would undo some of Black's accomplishments in establishing a widely agreed upon style guide.

These decisions above aren't set in stone, but they exist to provide a good user experience. We may be able to lift some of those restrictions once we're done with rule categorization or when Ruff supports plugins. But we have yet to start working on these projects. We could decide that this style guide is so essential that it is worth regressing the experience for some users. I'm not familiar enough with the scientific Python community to make that case, but you may have some data to demonstrate the importance.

That's why I recommend you consider a more incremental approach. Is there a limited set of rules that would significantly improve your experience and don't conflict with the formatter or existing lint rules? Are there changes we could make to Ruff's formatter style guide that would significantly improve your experience and that the majority of non-scientific users would also perceive as an improvement?

lucascolley:
thanks for the explanation @MichaReiser ! I'm not really sure what I think right now. maybe @jarrodmillman @stefanv have thoughts?

MichaReiser:
I’m sorry I made you feel that way. I reread my message, and I can see that I should have phrased it differently. It doesn’t give you a path forward; it only outlines the problem. I did phrase it so defensively because I can't promise that Ruff will be able to support the specification as is, at least not in the short-term.

A possible next step is to split this proposal into smaller pieces. For example, the style guide specifies when expressions should be parenthesized but also touches on operator spacing. Maybe we can even go more granular than this. 

I think we could probably implement a lint rule for parentheses handling without much problem because I believe it isn’t incompatible with the formatter or any existing lint rule. We could even explore if it’s worth implementing it as part of the formatter.

Getting the operator spacing into ruff will probably be more challenging because of the concerns that I outlined in my previous response. Maybe we can implement parts of it. Some parts may require more extended discussions and possibly compromises. There’s also the chance that we aren’t in a position to support some of it today.

To phrase it differently. The next step is to take this specification and make concrete proposals for how to extend (new lint rule, changed lint rule, changed formatting…) Ruff. I can try to help you with this.

tupui:
Sounds like we need to get this in to move forward. @mdhaber @lucascolley are you ok with that? Note that the process for SPECs is changing. We can, and should, get this in now and it would be marked as a draft at minima.

lucascolley (approved):
> Sounds like we need to get this in to move forward. @mdhaber @lucascolley are you ok with that? Note that the process for SPECs is changing. We can, and should, get this in now and it would be marked as a draft at minima.

👍

mdhaber:
> How do you identify mathematical expressions? Does it solely depend on the use of an operator (compare, binary, ...)?

I think it would depend solely on the use of the operators. In that example, the formatting would be the same with these rules as you've shown. There may be cases in which these rules would suggest funny things for overloaded operators, but my guess is that in most cases, it would just mean a `# noqa` here and there.

> One goal of Black's and Prettier's formatting style is to be predictable and easy to understand (to avoid cases where you're like: Uhhm, why did it do this?). That's why they try to avoid context-specific formatting where e.g. an expression gets formatted differently when embedded in a different context (exceptions apply).

Sure. There are costs to the simplicity, though. PEP8 explicitly recommends "If operators with different priorities are used, consider adding whitespace around the operators with the lowest priority(ies)". For the sake of simplicity (presumably), Black/[E225](https://docs.astral.sh/ruff/rules/missing-whitespace-around-operator/) just ignore this recommendation. The downside is that some find this makes the resulting expression more difficult to read. The goal of this SPEC is to capture the spirit of the PEP8 recommendation in rules. They will not be as simple as "add whitespace around all operators", but some may find that the benefits outweigh this cost.

> Have you looked at how other languages like Rust, Dart, JS, format mathematical expressions?

- Rust rules are [here](https://doc.rust-lang.org/nightly/style-guide/expressions.html#binary-operations). e.g. "Do include spaces around binary ops." "Use parentheses liberally; do not necessarily elide them due to precedence." 
- Dart rules are [here](https://github.com/dart-lang/dart_style/wiki/Formatting-Rules). e.g. "Spaces around binary and ternary operators." "Place binary operators on the preceding line in a multi-line expression." "No spaces around unary operators."
- JS rules are [here](https://google.github.io/styleguide/jsguide.html#formatting-horizontal-whitespace). e.g. "a single internal ASCII space also appears in the following places only...4. On both sides of any binary or ternary operator."

So, for the whitespace question, they all have essentially the same recommendation as Black and E225: always surround operators with whitespace. Others may have different goals, but one of my motivations here was to avoid that blanket rule. I think one of the reasons $LaTeX$ is preferred by many for mathematical documents is it goes to some lengths to make the whitespace look (subjectively) "good". We aren't (and can't, as we're assuming monospaced font) going that far here, but it's a small step in that direction.

> My biggest concern with Ruff's and Black's formatting is the nested expression...

Because of 8) "If line breaks must occur within a compound subexpression, the break should be placed before the operator with lowest priority." and "If there are multiple candidates, include the break at the first opportunity.", these (and some other variations) would satisfy the rules prescribed here:

```
(aaaaaaaaa
 + bbbbbbbbbbbb*ccccccccccccccc*ddddddddddddddd*xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
 + yyyyyyyyyyyy + zzzzzzzzzzzzzz)

(
 aaaaaaaaa
 + bbbbbbbbbbbb*ccccccccccccccc*ddddddddddddddd*xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
 + yyyyyyyyyyyy
 + zzzzzzzzzzzzzz
)
```

lucascolley:
> We can, and should, get this in now and it would be marked as a draft at minima.

Is this ready to go in as draft @stefanv ?

stefanv:
@lucascolley Thanks for the ping. Yes, it's definitely in good enough shape to shop around for discussion.

Note that there is one unclosed LaTeX expression, which would be good to fix first.

Some personal remarks:

- It feels like there could be more clarity between (1) and (5) around when brackets may be / should be inserted to improve an expression.
- `a + x*y**3` is rather hard to parse for my eyes, would probably have preferred `a + x * y**3`, but could be just me, and I suppose we'll get used to it. It's kind of funny to see how the pendulum swings. In the early days, everything was compressed: `a+x*y**3`, then spaces started creeping in, eventually expressions were fully spaced: `a + x * y ** 3`. Now the tide is rolling back again :)
- The definition section is a bit daunting, with all the descriptions of expressions, subexpressions, etc.; may be worth thinking about how to simplify that introduction, or move it to the end (most of the language is fairly intuitive, where it is used, so should be OK to have a glossary for reference?)
- Intro: "This leads to individual interpretation and styles which may conflict with those of others."; that sentence can perhaps be clarified along the lines of: "This leads to varying, even conflicting, mathematical expression styles across the ecosystem."

Anyone on here should feel free to move the PR out of draft status and merge, when they're ready. Or give me the thumbs-up and I'll do so.

lucascolley:
> a + x*y**3 is rather hard to parse for my eyes

I shared this concern, but after speaking to Matt I think it's fine considering the specification in (5) that there must be just a single `**` in the rightmost position if there is one. I think it's difficult to parse only because I'm used to expecting the operators to appear in any order, rather than having all expressions formatted in a standard way.

lucascolley:
Thanks again @MichaReiser for the points you raised above. I've had a little more time to think about them and I think you did a great job of explaining the challenges of adopting this in Ruff.

> - We decided not to implement rules that conflict with the formatter because formatted code should never raise new lint violations. This property is essential to integrate the formatter into check.
> - Some of the proposed style guide rules do conflict with the formatter.
The style guide conflicts with existing rules, and we can't just remove them. We have users who depend on them, even if the rules might go beyond what PEP-8 specifies. Arguably, we should never have added those rules,, but here we are.

One idea that I'm not sure has been considered yet would be a completely separate mode of the formatter + linter. The start point would be implementing this specification as lint rules and being able to format code to comply with it, but *in isolation* (just this spec). If that could be achieved, then any rules and formatting that do not conflict with this spec could then be added into this separate mode.

> - Implementing a formatting style guide is a multi-week, if not multi-month, project. It also increases the complexity to push any new formatting changes because they now need to fit into and be tested with multiple styles guides.

Of course, the main problem is always time. At least with this suggestion, the separate mode wouldn't block new formatting changes from integrating with "normal" ruff. That work would have to happen at some point if we wanted those changes in the separate mode, but in theory it could be done at a way later point.

> - 	I'm very hesitant about introducing a new style guide into the formatter because it would undo some of Black's accomplishments in establishing a widely agreed upon style guide.

Makes sense, although the context for this discussion is that Black’s accomplishments have been largely unable to reach the Scientific Python community due to concerns like this and the formatting of arrays. So we would at least be extending past Black’s accomplishments in some ways :)

Now I can see a lot of arguments for why a completely separate mode like this wouldn’t belong in Ruff’s API but be a separate package which uses Ruff’s machinery under the hood. Maybe that is what we’re looking for.

—-

Still, I think there would be value in translating the non-conflicting parts of this spec to additional lint rules. I’m just trying to be ambitious :)

spec-0012/index.md
mdhaber:
```suggestion
     e.g. $\frac{t}{v} \cdot \frac{x}{y}$.
```

mdhaber:
> Note that there is one unclosed LaTeX expression, which would be good to fix first.

Done.

> It feels like there could be more clarity between (1) and (5) around when brackets may be / should be inserted to improve an expression.

Can you give an example expression?

> a + x*y**3 is rather hard to parse for my eye

Perhaps you're happy with Black's rules for mathematical expressions, and that's fine. I got the sense that others weren't, though, so the rules here attempt to balance the conflicting desires:

- include whitespace, which makes it easier to distinguish multi-character variables from one another and from operators, and
- remove whitespace to visually reinforce the order of operations (as typeset math does and PEP8 calls for)

_without_ requiring the addition of lots of parentheses in expressions involving the common PEMDAS operators.

When I type `$a+x \cdot y^3$`, it renders as $a+x \cdot y^3$, adding more space around the addition than the multiplication, and none for the exponentiation. (And of course, we usually omit the $\cdot$ operator unless there's a reason to add it for emphasis, so $xy$ usually renders without any space.) There might be something to be said for including different amounts of whitespace for different priority operators, e.g. `x  +  y * x**3`, and I'd be happy to suggest a set of rules based on that. But assuming that's not in the cards and we only get one whitespace around operators, I preferred to use it to show that the multiplication happens before the addition even though the addition operation appears first on the line.

That's the rationale, but other viewpoints (e.g. "whitespace is only omitted around the highest-priority operator of a compound expression") are perfectly valid, too, so I don't think we'll be able to reason out which is best. I think it may be a matter of either adopting a mostly-baked proposal like this one or putting forth equally complete alternatives and voting. 

> The definition section is a bit daunting...

Ah, right, Pamphile mentioned that before, and we discussed moving it to a reference section at the end. I'll do that.

> Intro...

Yeah, I'll take another look at that. 

> And I think the rules don't forbid developers to include extra parentheses in cases like this if they would like to be extra clear.

The first rule is "do not add extraneous parentheses", but the last is "Any of the preceding rules may be broken if there is a clear reason to do so." So if the rules lead you to an expression that is particularly difficult to parse, break them. But `a + (x * y**3)` is not recommended because the rules would permit `a + x*y**3`.

(Another set of rules I considered would be based on "Whitespace is only omitted around the highest-priority operator of a compound expression" and something like "Expressions with more than two operator priority levels must make subexpressions with two operator priority levels explicit.* This would mean you would *have* to write `a + (x * y**3)`. But I didn't think people would want all those parentheses.)

lucascolley:
> The first rule is "do not add extraneous parentheses", but the last is "Any of the preceding rules may be broken if there is a clear reason to do so."

I guess I was alluding to the possibility that a formatter wouldn't add parentheses by default, but also wouldn't take them away if they are already there.

stefanv:
pre-commit.ci autofix

stefanv:
> * It feels like there could be more clarity between (1) and (5) around when brackets may be / should be inserted to improve an expression.

Sorry, that should have been (0) and (5).

It's not super intuitive to me when you can / should / must use brackets; but it may well be that the rules are comprehensive.

Wouldn't it be OK to always allow stylistic brackets, for aesthetic grouping, or grouping that is logical to the author?

[This discussion is not a blocker to merging the PR, btw.]

mdhaber:
Yes, see rule 9.

stefanv:
> Yes, see rule 9.

"Any of the preceding rules may be broken if there is a clear reason to do so."

mdhaber:
Yes. Because you asked:

> Wouldn't it be OK to always allow stylistic brackets, for aesthetic grouping, or grouping that is logical to the author?

It even shows examples of adding parentheses that aren't necessary.

stefanv:
I just posted the quote so others don't have to look it up.

mdhaber (commented):
@stefanv Math still didn't render. Any thoughts?

image

stefanv:
> @stefanv Math still didn't render. Any thoughts?
> 
> image

Requires a theme update. Fix has already been made. /cc @jarrodmillman

spec-0012/index.md
```diff
@@ -0,0 +1,202 @@
...
+2. Always surround non-PEMDAS[^2] operators with whitespace, and always make the priority of
+   non-PEMDAS operators explicit. For example, prefer `(x == y) or (w == t)` over
+   `x==y or w==t`.[^3]
+3. Always surround AS[^2] operators with whitespace.
+4. Typically, surround MD[^2] operators with whitespace, except in the following situations.
+   - When there are lower-priority operators (namely AS) within the same compound
+     expression[^4]. For example, prefer `z = -x * y**t` over `z = -x*y**t`, but
+     prefer `z = w + x*y**t` over `z = w + x * y**t` due to the presence of the
+     lower-priority addition operator.
+   - The division operation would be written mathematically as a fraction with a
```
mdhaber, line 49:
```suggestion
   - When the division operation would be written mathematically as a fraction with a
```

spec-0012/index.md
```diff
@@ -0,0 +1,202 @@
...
+1. Always use the `**` operator and unary `+`, `-`, and `~` operators _without_
+   surrounding whitespace. For example, prefer `-x**4` over `- (x ** 4)`.
+2. Always surround non-PEMDAS[^2] operators with whitespace, and always make the priority of
+   non-PEMDAS operators explicit. For example, prefer `(x == y) or (w == t)` over
+   `x==y or w==t`.[^3]
+3. Always surround AS[^2] operators with whitespace.
+4. Typically, surround MD[^2] operators with whitespace, except in the following situations.
+   - When there are lower-priority operators (namely AS) within the same compound
+     expression[^4]. For example, prefer `z = -x * y**t` over `z = -x*y**t`, but
+     prefer `z = w + x*y**t` over `z = w + x * y**t` due to the presence of the
```
mdhaber, line 47:
```suggestion
     prefer `z = w - x*y**t` over `z = w - x * y**t` due to the presence of the
```

spec-0012/index.md
```diff
@@ -0,0 +1,202 @@
...
+     operator due to the presence of the lower-priority addition operator. However,
+     this would lead to `t**v*x**y` being an implicit subexpression without spaces
+     containing more than one `**` operator. This code would be executed as
+     `(t**v)*(x**y) + z`, but the explicit parentheses should be included for clarity.
+   - In the expression `z + x**y/w`, no spaces are used around the division operator
+     due to the presence of the lower-priority addition operator. However, this would
+     lead to `x**y/w` being an implicit subexpression without spaces containing `**`
+     to the left of another operator. Options for refactoring include the addition of
+     parentheses (e.g. `z + (x**y)/w`) or pre-multiplying the exponential by a
+     fraction (i.e. `x + 1/w*x**y`).
```
mdhaber, line 80:
```suggestion
     to the left of another operator. This code would be executed as `z + (x**y)/w`),
     but the explicit parentheses should be included for clarity.
```

spec-0012/index.md
```diff
@@ -0,0 +1,201 @@
...
+[implied line continuation](https://peps.python.org/pep-0008/#maximum-line-length) and
+and [breaking lines before binary operators](https://peps.python.org/pep-0008/#should-a-line-break-before-or-after-a-binary-operator)[^1].
+
+0. Unless otherwise specified, rely on the implicit order of operations;
+   i.e., do not add extraneous parentheses. For example, prefer `u**v + y**z`
+   over `(u**v) + (y**z)`, and prefer `x + y + z` over `(x + y) + z`. A full
+   list of implicit operator priority levels is given by
+   [Operator Precedence](https://docs.python.org/3/reference/expressions.html#operator-precedence).
+1. Always use the `**` operator and unary `+`, `-`, and `~` operators _without_
+   surrounding whitespace. For example, prefer `-x**4` over `- (x ** 4)`.
```
mdhaber, line 39:
```suggestion
   surrounding whitespace. For example, prefer `y = -x**4` over `y = - (x ** 4)`.
```

spec-0012/index.md
```diff
@@ -0,0 +1,201 @@
...
+   surrounding whitespace. For example, prefer `y = -x**4` over `y = - (x ** 4)`.
+2. Always surround non-PEMDAS[^2] operators with whitespace, and always make the priority of
+   non-PEMDAS operators explicit. For example, prefer `(x == y) or (w == t)` over
+   `x==y or w==t`.[^3]
+3. Always surround AS[^2] operators with whitespace.
+4. Typically, surround MD[^2] operators with whitespace, except in the following situations.
+   - When there are lower-priority operators (namely AS) within the same compound
+     expression[^4]. For example, prefer `z = -x * y**t` over `z = -x*y**t`, but
+     prefer `z = w - x*y**t` over `z = w - x * y**t` due to the presence of the
+     lower-priority addition operator.
```
lucascolley, line 48:
```suggestion
     expression[^4]. For example, prefer `z = -x * y**t` over `z = -x*y**t`, but
     prefer `z = w - x*y**t` over `z = w - x * y**t` due to the presence of the
     lower-priority subtraction operator.
```

PR #344, opened 2024-09-11
Refresh SPEC 0 chart and schedule

jarrodmillman:


bsipocz (approved):
Are we doing this every month, or just quoterly (I was looking into the script yesterday, but then didn't commit the changes as we're still in Q3)?

jarrodmillman:
I was planning to do it quarterly, but then decided to do this one since Python 3.13 is coming out soon. Not sure it is necessary.

PR #345, opened 2024-09-11
Add skimage endorsements for SPECs 6, 7, and 8

jarrodmillman:
Discussed on skimage developer mailing list.

### DIFF
--- a/spec-0006/index.md
+++ b/spec-0006/index.md
@@ -9,6 +9,7 @@ author:
   - "Lars Grüter <lagru@mailbox.org>"
 discussion: https://discuss.scientific-python.org/t/spec-6-keys-to-the-castle
 endorsed-by:
+  - scikit-image
 ---
 
 ## Description
--- a/spec-0007/index.md
+++ b/spec-0007/index.md
@@ -10,6 +10,7 @@ author:
   - Other participants in the discussion <not.yet@named.org>"
 discussion: https://github.com/scipy/scipy/issues/14322
 endorsed-by:
+  - scikit-image
 ---
 
 ## Description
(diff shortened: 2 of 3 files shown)


stefanv (approved):

jarrodmillman:
Merging given that this was already approved on the mailing list.

lagru:
Hey you two, I'm a bit surprised that this has been merged already and within a time window of 1-2 hours. This is probably a misunderstanding about how I worded the [post on the mailing list](https://discuss.scientific-python.org/t/endorsing-specs-3-6-7-8/1412).

> Merging given that this was already approved on the mailing list.

I don't understand my post that way at all. I wrote

> We will open draft PRs in the next few days to endorse the SPECs and will notify you here. 

This would have given everyone another nudge to actually read each SPEC. I don't think any porcelain has been broken, but maybe we can wait a bit longer next time for project level stuff like this. :blush:

lagru (commented):
✅ Approving this PR now.

stefanv:
Hey Lars, since we had no responses on the post, and since this was cleared with you, Marianne, myself, and Jarrod, I thought it was OK to move ahead. Jarrod did ping me about it to check.

Now, I understand the caution, but we also have a problem that team members are often not very actively engaged. For big decisions, we need to make sure everyone is on board, but things like this feel like the answer is overwhelmingly likely to be "go ahead". Also, although there is no trend, nothing irreversible has been done, and a SPEC can be unendorsed. But, these specific SPECs encode the way we work, so that would seem odd.

I urge the team to operate in an optimistic fashion. Make a best guess as to what the others would say, and move ahead, unless the impact is likely to be large or irreversible. We have little momentum, and can't afford to lose any, with the little person-power available. If we make mistakes in our assessments, we make it right.

If you want this to be a broader discussion, I suggest we open a topic on the core mailing list so others will see it.

stefanv:
Noted, and thanks for sharing your concerns. We should encourage discussing those whenever they arise.

PR #350, opened 2024-10-03
SPEC-0007: rename validate to normalize

Carreau:
Valiate usually convey a notion of boolean or raising an error, here we are normalizing because we are rreturning potentially a different object.

### DIFF
--- a/spec-0007/transition_to_rng.py
+++ b/spec-0007/transition_to_rng.py
@@ -134,7 +134,7 @@ def _transition_to_rng(old_name, *, position_num=None, end_version=None):
                     message = (
                         f"Positional use of `{NEW_NAME}` (formerly known as "
                         f"`{old_name}`) is still allowed, but the behavior is "
-                        "changing: the argument will be validated using "
+                        "changing: the argument will be normalized using "
                         f"`np.random.default_rng` beginning in SciPy {end_version}, "
                         "and the resulting `Generator` will be used to generate "
                         "random numbers."
@@ -142,7 +142,7 @@ def _transition_to_rng(old_name, *, position_num=None, end_version=None):
                     warnings.warn(message, FutureWarning, stacklevel=2)
 
             elif as_new_kwarg:  # no warnings; this is the preferred use
-                # After the removal of the decorator, validation with
+                # After the removal of the decorator, normalized with
                 # np.random.default_rng will be done inside the decorated function
                 kwargs[NEW_NAME] = np.random.default_rng(kwargs[NEW_NAME])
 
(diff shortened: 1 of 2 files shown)


spec-0007/transition_to_rng.py
```diff
@@ -134,15 +134,15 @@ def wrapper(*args, **kwargs):
...
+                        "changing: the argument will be normalized using "
                         f"`np.random.default_rng` beginning in SciPy {end_version}, "
                         "and the resulting `Generator` will be used to generate "
                         "random numbers."
                     ) + cmn_msg
                     warnings.warn(message, FutureWarning, stacklevel=2)
 
             elif as_new_kwarg:  # no warnings; this is the preferred use
-                # After the removal of the decorator, validation with
+                # After the removal of the decorator, normalized with
```
mdhaber, line 145:
```suggestion
                # After the removal of the decorator, normalization with
```

mdhaber (approved):

stefanv:
The intent here was validate, not normalize, although I can see why that would be a sensible replacement. I mean, we're not trying to normalize the input to that function, we're trying to ensure that it is something that can be converted to rng.

Perhaps fully correct would be both: validate and normalize [the seed]"

mdhaber:
> I mean, we're not trying to normalize the input to that function, we're trying to ensure that it is something that can be converted to rng.

But `default_rng` is what we are suggesting to be called, and it does more than validate. It actually *does* convert to a `Generator`.

However, I was also thinking "validate and normalize", unless "normalize" implies "validate" when the input is already a `Generator`.

Carreau:
I agree with @mdhaber, and I think we do want to normalize, int/None and other are _convenience_, and any internal implementation should only allow `Generator`s, so _those_ should validate that arguments are generator. 



More generally you will find the saying "Parse don't validate" (I'll let you google it), and here it could apply because before `default_rng` the type id `Generator|None|int`, and after it's only `Generator`, and you do want as much as possible to be able to reason about the type.

(technically I don't even like `rng = np.random.default_rng(rng)`, I even think it should not have the same name after normalisation to reflect the change in type).


> unless "normalize" implies "validate"

I believe it does as it can't normalize an invalid value, but I'm happy to have both.

Carreau:
Actually going further, maybe the spec should say that user exposed functions should normalize (to a generator) and that private functions should validate (a generator only).

stefanv:
> Actually going further, maybe the spec should say that user exposed functions should normalize (to a generator) and that private functions should validate (a generator only).

That phrasing works for me.

mdhaber:
> Actually going further, maybe the spec should say that user exposed functions should normalize (to a generator) and that private functions should validate (a generator only).

I would rather just say what exposed functions do... normalize. Whether private functions need to validate is up to the library. I don't see why a private function should need to validate if all public functions normalize.

PR #352, opened 2024-10-04
Endorse spec 6 (Keys to the castle), by IPython

Carreau:
Also PyPI is upper case I, as it it Py.thon P.ackage I.ndex, and why it is pronouced /ˌpaɪpiˈaɪ/ and not /ˌpaɪˌpaɪ/ :-)


stefanv:
Thank you :)

PR #355, opened 2024-10-23
Sort projects

jarrodmillman:
This is a temporary fix. We should update the website to sort this for us.


jarrodmillman:
Self-merging as this is cosmetic and I plan to make PRs on top of these changes.

PR #364, opened 2024-11-13
NumPy project endorses SPEC 8

InessaPawson:
After a review and discussion with the maintainers and developer community, the NumPy Steering Council came to a consensus on endorsing SPEC 8

### DIFF
--- a/spec-0008/index.md
+++ b/spec-0008/index.md
@@ -12,6 +12,7 @@ author:
 discussion: https://discuss.scientific-python.org/t/spec-8-supply-chain-security
 endorsed-by:
   - networkx
+  - numpy
   - scikit-image
 ---
 


tupui:
@InessaPawson Woo this is great news for all these SPEC 👏 Are there public meeting notes? Otherwise could you get a few folks formally approving these PRs?

I am curious for this one specifically, does it mean you have plans to have continuous deployment as described in the SPEC?

seberg:
Right, happy to endorse it as well, it would definitely be nice to improve.  I think the NASA grant might do something about it, and if there are ideas on rephrasing things, then that can also happen at that time.

Also ping @rgommers.

InessaPawson:
>  Are there public meeting notes? 

@tupui We have discussed these SPECS on the NumPy mailing list: https://mail.python.org/archives/list/numpy-discussion@python.org/thread/KDP7EA2DBLZ7M7IQCPYWULI66H43NPKX/ and at 2 NumPy triage meetings, on [October 2](https://github.com/numpy/archive/blob/main/triage_meetings/2024/triage-2024-10-02.md) and [November 13](https://github.com/numpy/archive/blob/main/triage_meetings/2024/triage-2024-11-13.md).

tupui (approved):
For the reproducible build part, this was intentionally left out of this SPEC as it's still a challenging problem. Some people argue that the trail of evidence is enough, some want to get the same hash. There is a lot happening in that space (e.g. the last posts from Seth on his blog about SBOM) and we will certainly follow up on that.

Let's get this in otherwise since we got a few 👍 

Thanks again 🚀

stefanv:
Thanks, folks!

Ralf wrote:

> We can't use this exactly as is for NumPy, and I don't 100% agree with everything to the last letter, but overall these are good recommendations for the average scientific Python project, so happy to endorse.

I realize that community-wide consensus may not be achievable, but it would be good if we could have that between the authors and endorsers at least. Ralf, would you be willing to highlight any concerns / missing items, so the authors can think about how to address them?

PR #377, opened 2025-02-25
scikit-learn endorsment of SPEC6

glemaitre:
Scikit-learn is applying good practices of SPEC6 and is happily endorsing it.


tupui (approved):
Thanks 🙏

PR #378, opened 2025-02-25
scikit-learn endorsment of SPEC8

glemaitre:
All practices mentioned are reasonable and we (scikit-learn) are leaning towards them. So we can endorse already this SPEC8.


tupui (approved):
Thanks a lot for endorsing all these SPEC 🙏 Glad to see scikit-learn onboard 💪🎸🚀

PR #389, opened 2025-05-13
Add LTS to SPEC 0

jarrodmillman:


bsipocz (approved):
I really like phrasing this as LTS, it feels to be a nice extension and compromise!

I haven't looked in into the test failures, obviously the approval is not for any potentially relevant failures.

virchan:
I think what's outlined in LTS is consistent with what scikit-learn is doing in scikit-learn/scikit-learn#30888. 

So I would like to ping @lesteve and @lucascolley to see if they have any thoughts or input.

spec-0000/index.md
```diff
@@ -42,6 +42,16 @@ For instance, if a newer minimum version of a core package is needed by a projec
 which is not backported to older versions.
 {{< /admonition >}}
 
+{{< admonition note >}}
+Certain projects (e.g., projects that have more resources) may wish to provide long-term support (LTS) of an additional year.
+
+Specifically, for projects wishing to provide LTS we recommend that:
```
mdhaber, line 48: The first sentence is that all projects should adopt a common policy. This goes against that, so I think we should acknowledge it in some way. If we do not recommend one over the other, the first sentence might change to be "adopt one of two time-based poicies...". If we have a preference for the first but recognize the need for the other, we might acknowledge that here.

spec-0000/index.md
```diff
@@ -32,8 +32,8 @@ All versions refer to feature releases (i.e., Python 3.8.0, NumPy 1.19.0; not Py
 
 Specifically, we recommend that:
 
-1. Support for Python versions be dropped **3 years** after their initial release.
-2. Support for core package dependencies be dropped **2 years** after their initial release.
+1. Support for Python versions be dropped **3 years** after the next version of Python is released.
+2. Support for core package dependencies be dropped **2 years** after the next version of the package is released.
```
mdhaber, line 36: A potential ambiguity might be avoided by
```suggestion
2. Support for core dependency versions be dropped **2 years** after the next version of the dependency is released.
```

spec-0000/index.md
```diff
@@ -32,8 +32,8 @@ All versions refer to feature releases (i.e., Python 3.8.0, NumPy 1.19.0; not Py
...
 
 {{< admonition note >}}
```
mdhaber, line 38: If I understand the admonition correctly, the recommendation is to provide at least the indicated support window and not much more, but we recognize that there may be extreme circumstances in which support for a core dependency must be dropped early (e.g. to support a critical bug fix). Since it is always true that projects may endorse a SPEC as good practice but not follow it for any reason, I'm not sure that the admonition is needed unless it's particularly important to highlight this example.

mdhaber (commented):
Suggestions about the new text are inline.

If we were modifying other text:
- I don't understand the connection between the timeframe over which new releases support old dependency versions and the timeframe over which a given (feature) release will get bug fix releases. I think the SPEC would be more focused if it only mentions the former.
- The "Ecosystem Adoption" and "Implementation" sections are empty. Is that desirable?
- The background and motivation appears at the end, after the policy itself. Was that intentional?
- The SPEC only mentions limiting support duration as a motivation. Does it also recommend providing *at least* a certain amount of time - just not much more? 

but if the intent is to address only LTS here, feel free to hide as out of scope.

lucascolley (commented):
I agree with Matt's comments, but otherwise this looks like a useful change!

spec-0000/index.md
```diff
@@ -32,8 +32,8 @@ All versions refer to feature releases (i.e., Python 3.8.0, NumPy 1.19.0; not Py
 
 Specifically, we recommend that:
 
-1. Support for Python versions be dropped **3 years** after their initial release.
-2. Support for core package dependencies be dropped **2 years** after their initial release.
+1. Support for Python versions be dropped **3 years** after the next version of Python is released.
```
lesteve, line 35: I am a bit confused about this. This seems like one more year of support for the non-LTS recommendation but I may be missing some context ...

For Python (yearly release cadence), "3 years after the next version is released" means "4 years since the initial release".

The [changes in chart.md](https://github.com/scientific-python/specs/pull/389/files#diff-4216c03ca697c0458f16ef08d3b5d832d4fad4f34565f0aef03526ff1cf84e65) seem to agree with this "one more year of Python support".

mdhaber: Good point. When talking about this correction of the _logic_, we've discussed only support for core package dependencies like NumPy. Since the time between releases is typically 6 months, we've actually discussed that we could drop support for an old version 18 months after the next version release. Since the release cadence of Python is annual, we could keep the same nominal time by making this _two_ years after the next release and still keep the same nominal schedule.

So one option is to change these numbers to 1.5 and 2. We could make *both* 2 years for simplicity, but indeed, that would change the nominal support time.

stefanv:
Observations from discussions here at EuroSciPy:

- The "after the next release" phrasing is confusing.
- With the "after the next" change, doesn't this immediately grow the window, so that the LTS version is now *two* years longer?
- As a maintainer, it's useful to think in terms of "oldest supported version": i.e., "minimum version of Python should be a bit older than 3 yrs", then look up release dates and find that version of Python. Or, to phrase differently: at the point in time where you need the information, you need to look back, so the forward-looking formulation is confusing.
- May be worth formulating this both in terms of support windows and "today - 3 years, what was released" to match different intuitive approaches. Some people like the graphic approach as-is.
- Can we also visualize LTS on existing plot by shifting the red line?
- Stretch goal: would be nice to have a JS version that fetches release information from PyPi and generates the plot dynamically.

spec-0000/index.md
```diff
@@ -32,8 +32,8 @@ All versions refer to feature releases (i.e., Python 3.8.0, NumPy 1.19.0; not Py
...
+2. Support for core dependency versions be dropped **2 years** after the next version of the dependency is released.
```
Carreau: I think that is weird. Ipython releases every month, so that changes almost nothing. For a project that releases every year that extends support for a year. I don't think a project support should depends on the future release cadence that may or may not change. 

This also directly go against the principle of having a writing on the wall when dealing with companies to have a predictable end of life. I can see people getting presured "please don't release because we still need n-1 to be supported longer"

spec-0000/index.md
jarrodmillman: Removed. See #409.

PR #397, opened 2025-07-02
Sync SPEC 7 with scipy

jorenham:
ref: scipy/scipy#23276


jorenham:
CI failure seems unrelated: `/bin/bash: line 1: yaml2ics: command not found`

bsipocz (commented):
The code itself looks reasonable, but this made me open #399 to question if we should have this code in the SPEC doc to begin with.

(and as the discussion in the scipy thread shows, it's not obvious if the preference is to keep or drop the keywords from the signature.)

PR #422, opened 2026-05-20
Add Astropy to core projects list

eteq:
This PR adds Astropy as a core project, as per discussion with @jarrodmillman 

cc @nstarman @ceb8 @aaryapatil @kbwestfall (the other current astropy coordinators)

PR #424, opened 2026-05-28
Fix licence URL

bsipocz:
The linkchecker didn't complain in #422, but it did when the update got pulled into the main website build.


stefanv:
It works 😊

stefanv (approved):

PR #433, opened 2026-09-21
SPEC 1: mention PEP 810 and plans for sunsetting the SPEC

neutrinoceros:
Close #413
not sure about the exact phrasing or formatting, but this looks like the gist of it.


lucascolley (approved):

bsipocz:
@jarrodmillman had a PR in work to update the SPEC1 text, I would be very much in favour of having that one in than this one.

